PR #35, opened 2021-04-27
Common exporter implementation. Merging ERC20 and ETH transfers.

YordanPavlov:
This PR is a proposal for a common 'San exporter' architecture. Currently the ERC20 and ETH transfers exporters are merged. If we are happy with the design we may move all the code to the san-exporter repository and deprecate the other.

The implementation tries to organize the code around a 'base' class called WorkerBase which is to be inherited by different blockchains. The base part of the code handles connection with Kafka, Zookeeper, metrics and do a work loop. Different implementations organize the export around a work() method which is to being invoked at intervals. In that method they set how much progress has been made, last primary key used, when to be invoked again and so on, using the member variables of the base class.


blockchains/eth/eth_worker.js
```diff
@@ -0,0 +1,254 @@
...
+
+    logger.log(`Fetching receipts of ${fromBlock}:${toBlock}`)
+    const receipts = await fetchReceipts(blocks)
+
+    blocks.forEach((block) => {
+      block.transactions.forEach((transaction) =>
+        result.push({
+          from: transaction.from,
+          to: block.miner,
+          value: parseFloat(web3.utils.hexToNumberString(transaction.gasPrice)) * parseFloat(web3.utils.hexToNumberString(receipts[transaction.hash].gasUsed)),
```
kalkov, line 168: Hard to read, better extract to a function and give meaningful name.

blockchains/eth/eth_worker.js
```diff
@@ -0,0 +1,254 @@
...
+  decodeTransferTrace(trace, blocks) {
+    const web3 = this.web3
+    const timestamp = web3.utils.hexToNumber(blocks.get(trace["blockNumber"]).timestamp)
+
+    // Block & uncle rewards
+    if (trace["type"] == "reward") {
+      return {
+        from: `mining_${trace["action"]["rewardType"]}`,
+        to: trace["action"]["author"],
+        value: parseFloat(web3.utils.hexToNumberString(trace["action"]["value"])),
```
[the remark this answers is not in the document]
kalkov: These small functions will be great candidates for unit tests.

docker/docker-compose.yml
```diff
@@ -43,11 +43,14 @@ services:
       BLOCK_INTERVAL: "50"
       EXPORT_TIMEOUT_MLS: 300000
       EXACT_CONTRACT_MODE: "1"
+      # Should be substituted from run.sh script with different values.
+      BLOCKCHAIN: ${BLOCKCHAIN}
```
kalkov, line 47: Wouldn't that make the file unusable w/o `run.sh`?

index.js
```diff
@@ -1,120 +1,118 @@
...
-  setTimeout(work, 30 * 1000)
-}
+  async setWorker() {
+    if (this.worker != null) {
+      throw new Error("Worker is already set")
+    }
 
-async function initLastProcessedBlock() {
-  const lastPosition = await exporter.getLastPosition()
+    switch (process.env.BLOCKCHAIN) {
```
kalkov, line 71: You can make worker initialization dynamic:
```js 
const WORKERS = { ERC20Worker, ERC20Worker }

export default function newWorker (name) {
  return WORKERS[name];
}
```
and then:
```js
const workerName = process.env.BLOCKCHAIN.toUpperCase()
this.worker = newWorker(workerName) 
```

kalkov:
I would put data files - csv, json in separate folders called `data_files` or something similar

test/eth/index.spec.js
```diff
@@ -0,0 +1,40 @@
+const assert = require("assert");
+const { addGenesisTransfers } = require("../../blockchains/eth/lib/genesis_transfers")
+const Web3 = require('web3')
+const web3 = new Web3()
+
+describe('genesis transfers', function() {
```
kalkov, line 6: Nice tests!

bin/run.sh
```diff
@@ -1,3 +1,3 @@
  #! /bin/sh
 
-docker-compose -f docker/docker-compose.yml up --build && docker-compose -f docker/docker-compose.yml rm -f
+BLOCKCHAIN=eth docker-compose -f docker/docker-compose.yml up --build && docker-compose -f docker/docker-compose.yml rm -f
```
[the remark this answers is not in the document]
YordanPavlov: Which part? We need the "rm -f" so that Kafka and Zookeeper containers are destroyed and not reused between restarts. And we need to provide the file name because it is not in the current directory.

blockchains/eth/eth_worker.js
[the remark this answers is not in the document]
YordanPavlov: Agree. However this is not the essence of the PR. What I have done is took the code from:
https://github.com/santiment/eth-transfers-exporter
and copy-pasted it here in a sort of validation that a common architecture is possible.

docker/docker-compose.yml
YordanPavlov: Yes, do we have another use case of the file? What other way we have for providing dynamic variable?

index.js
YordanPavlov: I am trying to weight in the pros of this syntax, do you find it cleaner? Also what do you mean by 'dynamic' am I missing some initialization difference?

test/eth/index.spec.js
YordanPavlov: Can't take any credit for those - just a copy-paste.

blockchains/eth/eth_worker.js
```diff
@@ -0,0 +1,254 @@
...
+        valueExactBase36: web3.utils.toBN(trace["action"]["value"]).toString(36),
```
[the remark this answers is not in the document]
YordanPavlov: Very similar exporters like ERC20 and ETH would share a lot of code, we can move that to a common utils file. In the common case however there would be no reuse of code under different /blockchains/* directories.

blockchains/eth/eth_worker.js
```diff
@@ -0,0 +1,254 @@
...
+    if (trace["type"] == "create") {
+      return {
+        from: trace["action"]["from"],
+        to: trace["result"]["address"],
+        value: parseFloat(web3.utils.hexToNumberString(trace["action"]["value"])),
+        valueExactBase36: web3.utils.toBN(trace["action"]["value"]).toString(36),
+        blockNumber: trace["blockNumber"],
+        timestamp: timestamp,
+        transactionHash: trace["transactionHash"],
+        transactionPosition: web3.utils.hexToNumber(trace["transactionPosition"]),
```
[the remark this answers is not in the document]
YordanPavlov: What would you suggest be extracted to a function, the arguments provided to `web3` are rather different?

blockchains/eth/eth_worker.js
YordanPavlov: Done.

blockchains/eth/eth_worker.js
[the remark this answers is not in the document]
kalkov: The DRY principle states: 
> Every piece of knowledge must have a single, unambiguous, authoritative representation within a system

I don't try to follow it blindly however if something is repeated three times it fires an alarm in my head and try to make it reusable. 
In the current situation that line is repeated four times: `web3.utils.toBN(trace["action"]["value"]).toString(36)`

Extracting to a function in the same file: 
```python 
def parse_value_exact_base_36(field):
  return web3.utils.toBN(field).toString(36)
end
```
will have the following benefits:
1) Will be more readable. When I am scanning a file and start to feel that lines are similar I try to find the differences  between them. 
Here the difference is clear(the field):
 ```python
valueExactBase36: parse_value_exact_base_36(trace["action"]["value"]) 
```
2) It will be a lot easier to change
3) It will be a lot easier to write unit test for that function. In this case it will be very handy to write such test - it will document the input and the output of the function.

File: blockchains/eth/eth_worker.js (hunk 8 above)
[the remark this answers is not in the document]
Comment (kalkov) [reply blockchains/eth/eth_worker.js:65]: Same as above

File: docker/docker-compose.yml (hunk 3 above)
Comment (kalkov) [reply docker/docker-compose.yml:47]: You can set a default value if the env var is not present:
```yaml
BLOCKCHAIN: ${BLOCKCHAIN:-ethereum}
```
https://docs.docker.com/compose/compose-file/compose-file-v3/#variable-substitution

index.js
kalkov: On a second thought - we won't add any new workers here in the future?

YordanPavlov: The plan is to add new workers from other blockchains. I think I got your idea now, by using the syntax you propose one would need to include a new worker only as 'require' and not edit the code. Done similarly to how you proposed but with a 'new' operator.

docker/docker-compose.yml
YordanPavlov: Default value provided in docker-compose.

blockchains/eth/eth_worker.js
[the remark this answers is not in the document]
YordanPavlov: Similar code is separated into a method.

blockchains/eth/eth_worker.js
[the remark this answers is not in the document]
YordanPavlov: OK i'm sold. Once I started doing it - it turned out that there are only four unique types of invocations. Now moved into four helper methods.

YordanPavlov:
> I would put data files - csv, json in separate folders called `data_files` or something similar

Let's leave that for now. There is a file '`dao_hack.js`' which is mostly data but with some code. I don't feel like refactoring that right now.

index.js
kalkov: Yep, that was the idea.

YordanPavlov: I've taken the idea up a notch. I dynamically use the blockchain name to load just the required file. This helps us not need contract mapping files for the ETH deploy for example.

kalkov (approved):

YordanPavlov:
A bug was found after deploying on stage. It was the result of some refactoring being done as we moved the codebase from the ETH transfers repo to this one. The result of the bug was producing timestamps in hex format. This broke the consumption of the data.

kalkov (approved):

PR #39, opened 2021-07-20
Copy san-exporter code instead of using it as dependency.

YordanPavlov:
A PR to move the san-exporter code inside this common exporter.

### DIFF
--- a/lib/kafka_storage.js
+++ b/lib/kafka_storage.js
@@ -1,0 +1,224 @@
+const Kafka = require("@santiment-network/node-rdkafka");
+const zk = require("node-zookeeper-client-async");
+const { logger } = require('./logger')
+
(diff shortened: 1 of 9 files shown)


lib/kafka_storage.js
```diff
@@ -0,0 +1,224 @@
...
+    if (transactional) {
+        producer_settings['transactional.id'] = this.topic_name;
+        producer_settings['enable.idempotence'] = true;
+    }
+
+    this.producer = new Kafka.Producer(producer_settings);
+    this.zookeeperClient = zk.createAsyncClient(ZOOKEEPER_URL);
+  }
+
+  get topic_name() {
```
kalkov, line 50: Should be camelCased

kalkov (approved):

lib/kafka_storage.js
YordanPavlov: Fixed.

PR #42, opened 2021-09-02
Add transfer exporter for matic

antisaa:


blockchains/matic/lib/fetch_events.js
```diff
@@ -0,0 +1,58 @@
+"use strict";
+const { getBlockTimestamp, decodeEventBasicInfo, decodeEvents } = require('../../erc20/lib/fetch_events')
+const { decodeAddress } = require('../../erc20/lib/util')
+
+const ZERO_ADDRESS = "0x0000000000000000000000000000000000000000"
```
YordanPavlov, line 5: Some of those are unused, let's remove them.

YordanPavlov (approved):
LGTM. What is lacking is extending the unit tests for the new type of 'matic' events. I guess we can add them later if we are in a hurry now.

PR #52, opened 2021-11-29
Change post-London fee logic so that all trx types burn fees

YordanPavlov:
A PR to fix the fee burning logic.
1. Even old style transactions burn the base fee after the London fork.
2. For the new type 2 transactions check if maxFeePerGas is not reducing the fee paid.

### DIFF
--- a/blockchains/eth/lib/constants.js
+++ b/blockchains/eth/lib/constants.js
@@ -3,11 +3,13 @@ const CONFIRMATIONS = parseInt(process.env.CONFIRMATIONS || "3")
 const PARITY_NODE = process.env.PARITY_URL || "http://localhost:8545/"
 const LOOP_INTERVAL_CURRENT_MODE_SEC = parseInt(process.env.LOOP_INTERVAL_CURRENT_MODE_SEC || "30")
 const BURN_ADDRESS = "burn"
+const LONDON_FORK_BLOCK = 12965001
 
 module.exports = {
     BLOCK_INTERVAL,
     CONFIRMATIONS,
     PARITY_NODE,
     LOOP_INTERVAL_CURRENT_MODE_SEC,
-    BURN_ADDRESS
+    BURN_ADDRESS,
+    LONDON_FORK_BLOCK
 }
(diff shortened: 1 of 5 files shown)


blockchains/eth/lib/constants.js
```diff
@@ -3,11 +3,13 @@ const CONFIRMATIONS = parseInt(process.env.CONFIRMATIONS || "3")
 const PARITY_NODE = process.env.PARITY_URL || "http://localhost:8545/"
 const LOOP_INTERVAL_CURRENT_MODE_SEC = parseInt(process.env.LOOP_INTERVAL_CURRENT_MODE_SEC || "30")
 const BURN_ADDRESS = "burn"
+const LONDON_FORK_BLOCK = 12965001
```
nailkhaf, line 6: why not 12965000?
https://etherscan.io/block/12965000

blockchains/eth/lib/fees_decoder.js
```diff
@@ -12,52 +11,114 @@ class FeesDecoder {
...
+     *
+     * EIP1559 transactions must specify both maxPriorityFeePerGas and maxFeePerGas. They must not specify gasPrice.
+     *
+     * https://besu.hyperledger.org/en/stable/Concepts/Transactions/Transaction-Types/
+     */
+    const maxPriorityFeePerGas = this.web3Wrapper.parseValueToBN(transaction['maxPriorityFeePerGas'])
+    const maxFeePerGas = this.web3Wrapper.parseValueToBN(transaction['maxFeePerGas'])
+    const baseFeePerGas = this.web3Wrapper.parseValueToBN(block.baseFeePerGas)
+
+    const minerFeePerGas = baseFeePerGas.add(maxPriorityFeePerGas).gt(maxFeePerGas) ?
```
nailkhaf, line 50: `transaction['gasPrice'] - block.baseFeePerGas`

When I query mined tx type 2 I got field gasPrice, isn't this the same?

YordanPavlov: Thanks for this observation. I did not read the whole EIP and thought that the 'gasPrice' field is a leftover to be removed. You are right that it must be present and we can depend on it.

YordanPavlov (commented):
Thanks for your input. I've addressed the recommendations.

nailkhaf (approved):

PR #65, opened 2022-03-09
Cardano genesis transfers

YordanPavlov:
Fixes on the Cardano exporter.

* Export the genesis transfers. The block which includes them do not have a number so our query missed them. We need to explicitly specify a block hash.
* Other fixes.

### DIFF
--- a/blockchains/cardano/cardano_worker.js
+++ b/blockchains/cardano/cardano_worker.js
@@ -66,8 +66,6 @@ class CardanoWorker extends BaseWorker {
             address
             value
           }
-
-
         }
       }
     `)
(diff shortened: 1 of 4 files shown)


blockchains/cardano/cardano_worker.js
```diff
@@ -36,13 +36,84 @@ class CardanoWorker extends BaseWorker {
...
+          inputs {
+            address
+            value
+          }
+
+          outputs {
+            address
+            value
+          }
+
```
antisaa, line 69: Are two empty rows intentional?

antisaa (approved):

PR #75, opened 2022-05-10
BNB exporter integration

YordanPavlov:
An integration of [BNB exporter](https://github.com/santiment/bnb-exporter) into the common san-chain-exporter. Along the way:
* Simplification of the BNB time interval logic
* Fix for a bug when `start interval > end interval`
* Tests added


tzanko-matev:
This is too big and it is impossible for me to review. It seems that in addition to adding BNB there are also other changes that are being made. Is it possible make separate PRs of each of those changes and merge them first?

tzanko-matev (approved):

PR #77, opened 2022-05-11
Enable running unit tests outside of docker

tzanko-matev:
There were two custom configurations done in Dockerfile-test:
1) copy of contract_mapping.json
2) set up of CONTRACT_MODE environment

Those customizations make it impossible to just run `npm test` outside
of docker.

We have made the following changes:

1) Added an environment variable which sets the path of the
contract_mapping.json file
2) Changed the `npm test` command to have the proper environment
variables set up
3) Removed the customizations from Dockerfile-test

As a result now it is possible to run `npm test` outside of Docker. On
my machine time of running unit tests went down as follows:

- run `./bin/test.sh` when no code changes were made - 4 seconds
- run `./bin/test.sh` when code changes were made - 7-8 seconds
- run `npm test` - 1.5 seconds

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,13 +21,24 @@ If the health check does not pass you get response code 500 and a message descri
 
 ## Tests
 
-You can run the unit tests with:
+You can run the unit tests in one of the following ways:
 
+1) Without Docker
+You need to have `nodejs` installed locally.
+```bash
+$ npm install
+$ npm test
+```
+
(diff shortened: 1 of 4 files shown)


YordanPavlov (approved):

spiderjako (approved):
Works just great, though not sure whether the results should be logged

PR #83, opened 2022-05-11
Refactor the common exporter code

YordanPavlov:
In this PR we refactor a part of the common exporter code. We move part of the logic from `index.js` to the base class we use for workers (`WorkerBase`). The code being there allows for specific worker implementations to overwrite it with specific logic. This refactor is intended to facilitate the introduction of BNB exporter. In this blockchain we fetch an API using time intervals instead of blocks, so part of the current presumptions were not valid anymore.

### DIFF
--- a/lib/worker_base.js
+++ b/lib/worker_base.js
@@ -1,4 +1,5 @@
 "use strict";
+const { logger } = require('./logger')
 
 class WorkerBase {
   constructor() {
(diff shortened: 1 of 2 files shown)


lib/worker_base.js
```diff
@@ -24,6 +25,51 @@ class WorkerBase {
...
   }
+
+  /**
+   * Should be overriden depending on blockchain implementation.
+   *
+   * @returns Number of new requests made towards the Node endpoint. Used for metrics purposes.
+   */
+  getNewRequestsCount() {
+    return 1
+  }
```
tzanko-matev, line 36: Shouldn't this be 0 in the base class?

antisaa (approved):

lib/worker_base.js
YordanPavlov: The presumption is that on each 'work loop' the Full Node would be queried at least once, that's true for all blockchains. I think 1 is less wrong that 0, but I guess I can throw exception in the base class and force overwrite. What do you think?

tzanko-matev (approved):

PR #87, opened 2022-05-13
Fix BNB local run

YordanPavlov:
* The NODE_URL for BNB needs to be replaced by run script
* Provide a more recent BNB start time so that more data is available


spiderjako (approved):
Looked through the files, seems okay. If tested out locally and works just fine, LGTM

PR #91, opened 2022-06-24
Migrate bnb trades

YordanPavlov:
A migration of [BNB trades exporter](https://github.com/santiment/bnb-exporter-trades) into the common exporter repository. A lot of code was reused from the already migrated [BNB transactions exporter](https://github.com/santiment/bnb-exporter).

### DIFF
--- a/bin/run_bnb_trades.sh
+++ b/bin/run_bnb_trades.sh
@@ -1,4 +1,6 @@
  #! /bin/sh
+set -e
 
 BLOCKCHAIN=bnb NODE_URL=https://dex.binance.org/api/v1/trades/ BNB_MODE=trades \
-docker-compose -f docker/docker-compose.yaml up --build && docker-compose -f docker/docker-compose.yaml rm -f
+docker-compose -f docker/docker-compose.yaml up --build
+docker-compose -f docker/docker-compose.yaml rm -f
--- a/bin/run_bnb_trades.sh
+++ b/bin/run_bnb_trades.sh
@@ -1,0 +1,4 @@
+ #! /bin/sh
+
+BLOCKCHAIN=bnb NODE_URL=https://dex.binance.org/api/v1/trades/ BNB_MODE=trades \
+docker-compose -f docker/docker-compose.yaml up --build && docker-compose -f docker/docker-compose.yaml rm -f
--- a/blockchains/bnb/lib/bnb_transactions_fetcher.js
+++ b/blockchains/bnb/lib/bnb_transactions_fetcher.js
@@ -1,6 +1,6 @@
 "use strict";
 const { logger } = require('../../../lib/logger')
-const {getLastBlockTimestamp} = require('./utils')
+const utils = require('./utils')
 const {fetchTimeInterval} = require('./fetch_transactions')
 const constants = require("./constants")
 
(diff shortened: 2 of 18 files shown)


apozdeyev (approved):

bin/run_bnb_trades.sh
```diff
@@ -0,0 +1,4 @@
+ #! /bin/sh
+
+BLOCKCHAIN=bnb NODE_URL=https://dex.binance.org/api/v1/trades/ BNB_MODE=trades \
+docker-compose -f docker/docker-compose.yaml up --build && docker-compose -f docker/docker-compose.yaml rm -f
```
WonderBeat, line 4: what do you think of `set -e`. So you can avoid `&&` here

blockchains/bnb/lib/bnb_transactions_fetcher.js
```diff
@@ -59,7 +60,11 @@ class BNBTransactionsFetcher {
       return nextRange
     }
 
-    this.lastBlockTimestamp = await getLastBlockTimestamp(metrics);
+    this.lastBlockTimestamp = this.bnbTradesMode ?
+    await getLastTradesBlockTimestamp(metrics)
+    :
```
WonderBeat, line 65: Dangling `:` looks strange. Can you pad it a little bit?
And btw, do you use any prettyfier in JS?

blockchains/bnb/lib/fetch_transactions.js
```diff
@@ -29,21 +43,35 @@ function sendTrxQuery(trxId, queue, metrics) {
...
-async function fetchTimeInterval(queue, startTimeMsec, endTimeMsec, nodeResponsePromises, metrics) {
+function getNumPagesToIterate(firstPageResult, startTimeMsec, endTimeMsec, bnbTradesMode) {
+  if (typeof firstPageResult === 'undefined' ||
+   (bnbTradesMode && firstPageResult.trade === 'undefined') ||
+   (!bnbTradesMode && firstPageResult.txArray === 'undefined')) {
+      throw ("Error in fetch interval ", startTimeMsec, " - ", endTimeMsec)
+  }
+
+  return bnbTradesMode ?
+  Math.ceil(firstPageResult.total / constants.NUM_TRADE_ROWS_FETCH)
```
WonderBeat, line 54: This additional parameter that we pass everywhere, oh, I wish we can avoid it. 
Can this switch be expressed with OOP somehow?
And this function in general? Why does it has so many arguments if it requires only one. The rest of params are only used to check preconditions, right? Can we check preconditions in another place? Current function signature confuses me a bit. It should "fetch time interval", but actually it just calculates the number of 'pages' in `firstPageResult`

blockchains/bnb/lib/fetch_transactions.js
```diff
@@ -188,9 +216,29 @@ function filterRepeatedTransactions(listTrx) {
   }
 }
 
+/**
+ * For reasons currently unknown the BNB API returns same results on different pages. Last result from page x may
+ * appear as first result on page x + 1. That is the oldest result on page x may be the newest on page x + 1.
+ * @param {*} listTrx An array of trades to be filtered
+
+ */
```
WonderBeat, line 224: Do you have a testcase for this?
Does it only work for sequential duplicates?

WonderBeat (commented):
LGTM, though this additional parameters increases overal code complexity

blockchains/bnb/lib/fetch_transactions.js
YordanPavlov: The function is stateless even though the arguments are indeed a lot. I am not sure OOP is simpler to grasp compared to procedural programming. A function that has no state and acts only upon arguments, even if they are more than 2 as per OOP rules is fine, no? I have broken it into a `check` part and main computation to simplify the logic.

bin/run_bnb_trades.sh
YordanPavlov: Sounds good. Done.

WonderBeat:
:+1:

PR #94, opened 2022-09-05
Enforce consistent styling in the codebase

spiderjako:
### Description
------
Main points:
* Changing to a more consistent styling all at once
* Adding a receipts worker, which fetches the data and puts it into kafka, functionality that is currently living in [eth-receipts-exporter](https://github.com/santiment/eth-receipts-exporter/), the idea being that we should centralise this functionality - EDIT: This has been moved to [another PR](https://github.com/santiment/san-chain-exporter/pull/101)
* While polishing the files in point one, I stumbled upon a wrongly initialised test object in the cardano worker's test, small thing, but still important to have everything working properly


YordanPavlov (commented):
I think the comment needs update:
```
Adding a receipts worker, which fetches the data and...
```
is no longer relevant. Also what is this coding standard based on? Something like this:
https://google.github.io/styleguide/jsguide.html
?

YordanPavlov (approved):

PR #100, opened 2022-09-28
Add versioning to ETH exporter

YordanPavlov:
Add a version to the output data of the ETH exporter.


YordanPavlov:
We have other ideas on how to recompute metrics around S3 data lake

PR #101, opened 2022-10-03
Add receipts functionality

spiderjako:
Migration of the [eth-receipts-exporter](https://github.com/santiment/eth-receipts-exporter) codebase to san-chain-exporter.

### DIFF
--- a/bin/run_receipts.sh
+++ b/bin/run_receipts.sh
@@ -1,0 +1,5 @@
+#! /bin/sh
+set -e
+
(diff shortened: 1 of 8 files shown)


YordanPavlov (commented):
Let's not loose unit tests along the migration, please migrate them as few as they are.

YordanPavlov (commented):
Also please always add a description. For example:
"This is a migration of this code base [link] to the common exporter."

YordanPavlov (approved):
LGTM but please do the output comparison test with the old repo before merge.

PR #111, opened 2022-11-30
Ability to run code on host machine and not inside Docker

YordanPavlov:
Introduce the ability to run the JS code on the host machine and not inside Docker. This eases debugging.


docker/docker-compose-servers.yaml
```diff
@@ -0,0 +1,40 @@
+version: "3.7"
+services:
+  zookeeper:
+    image: confluentinc/cp-zookeeper
+    #logging:
```
spiderjako, line 5: Why do we comment out the logging-driver line instead of just removing it

spiderjako (approved):
LGTM, left minor comments

docker/docker-compose-servers.yaml
YordanPavlov: Damn, I've commited a lot of local changes I used for testing, sorry about this. Should be reverted now.

spiderjako (approved):

PR #112, opened 2022-11-30
Copy filter_errors.js file from blockchain-utils repo

YordanPavlov:
This PR copies the only file in the [repo](https://github.com/santiment/blockchain-utils). I think having a repo with a single utility file doesn't make sense. Moreover I have identified that this filter function works **_very_** slow for some blocks, starving the Node thread for 20-30 seconds. I would fix this in another PR.

As next step we can deprecate the `blockchain-utils` repo.

### DIFF
--- a/blockchains/eth/lib/filter_errors.js
+++ b/blockchains/eth/lib/filter_errors.js
@@ -1,7 +1,28 @@
 const trie = require('trie-prefix-tree')
 const { groupBy } = require('lodash')
 
+/**
+ * A function to mark children traces as erroneous if their parent is an error.
+ *
+ * To achieve this we depend on the `traceAddress` field. The `traceAddress` field of all returned traces,
+ * gives the exact location in the call trace [index in root, index in first CALL, index in second CALL, …].
+ * i.e. trace with traceAddress for this call on the right
+ *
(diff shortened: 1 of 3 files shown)


blockchains/eth/lib/filter_errors.js
```diff
@@ -0,0 +1,41 @@
...
+      if (errAddresses.indexOf('-1 ' + trace.traceAddress.join(' ')) !== -1) {
+        trace['error'] = 'parent_error'
+      }
+    })
+  })
+  return [].concat(...Object.values(txs))
+
+}
+
+function filterErrors(traces) {
```
spiderjako, line 33: https://github.com/santiment/blockchain-utils/blob/master/eth/filter_errors.js#L62
Is the `filterNonActions` function removed on purpose? I remember there was some drama and you added that check
https://github.com/santiment/blockchain-utils/commit/9761ff7b11c889aad2942b1ba47294bc4d705558

YordanPavlov: Oh my... I had copied an older version of the file. Thank you for noticing this!

spiderjako (approved):

PR #114, opened 2022-12-07
Filter errors starve fix

YordanPavlov:
Custom Trie implementation which drastically reduces the time it takes to mark failed traces. [The problem](https://www.notion.so/santiment/ETH-transfers-exporter-stuck-13254a4b3bad426bb69a7b1926f822ee) was initially discovered by exporter timeouting connection to Zookeeper. Since the Node thread was busy for 20+ seconds working on ETH traces, the process could not send Zookeeper heartbeat messages.

In the previous implementation we used the [trie-prefix-trie](https://github.com/lyndseybrowning/trie-prefix-tree) library for this purpose. Using it, we did the following steps:

1. Build the trie. Complexity O(N). Where N is number of traces. We may be more specific and count this as O(N*M) where M is length of error stack ( that is this field  `{"traceAddress": [ 0, 0, 1]}` inside each trace object. However let's simplify.
2. Go over traces and filter error ones, for each such trace find its children. Complexity is O(N* log(N)).
3. For all failed children go over the original traces list and mark them as failed. Complexity is O(N*N).

So this algorithm has an overall complexity of O(N*N).

On the other hand, what we can do with a custom Trie structure is:

1. Build the trie. O(N).
2. Mark all failed children. O(N).

This gives us the much better complexity of O(N).

Note:
* Most of the test in `test/eth/filter_errors/filter_errors.spec.js` were copied from the previous [repo](https://github.com/santiment/blockchain-utils/blob/master/test/index.spec.js). Forgot to do it on [copying the code](https://github.com/santiment/san-chain-exporter/pull/112).
* A new test is added to illustrate the performance problem: `describe('Trie performance for large trace')`


apozdeyev (approved):
👍

spiderjako (approved):

PR #120, opened 2022-12-29
Fixing issue with Kafka keys

spiderjako:
## Description
The short backstory is that I've ran the exporter for litecoin, but the problem is, that the "primaryKey" key is non-existent in the fetched data from the API, hence it pushed the data in Kafka with a primaryKey which was null and it broke the whole ltc balances pipeline on stage

### DIFF
--- a/lib/store_events.js
+++ b/lib/store_events.js
@@ -1,18 +1,17 @@
 'use strict';
 
 const { logger } = require('./logger');
+const { BLOCKCHAIN } = require('./constants');
 
 async function storeEvents(exporter, events) {
   await exporter.beginTransaction();
-
   try {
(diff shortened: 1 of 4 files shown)


lib/store_events.js
```diff
@@ -8,6 +8,8 @@ async function storeEvents(exporter, events) {
   try {
     if (exporter.exporter_name === 'bsc_receipts_exporter') {
       await exporter.sendDataWithKey(events, 'transactionHash');
+    } else if (exporter.exporter_name === 'utxo-litecoin-v2') {
```
YordanPavlov, line 11: We want to have this logic for all UTXO based blockchains, so let's use the `BLOCKCHAIN` env variable as distinguisher.

YordanPavlov (approved):
LGTM

PR #123, opened 2023-01-12
Share logic for next interval calculation

YordanPavlov:
* Move logic for next interval calculation into a function. Reuse it.
* If Node goes backwards, this should not break the interval logic, just do not progress.

### DIFF
--- a/blockchains/eth/lib/next_interval_calculator.js
+++ b/blockchains/eth/lib/next_interval_calculator.js
@@ -1,5 +1,9 @@
 const constants = require('./constants');
 
+
+function isFullIntervalAvailable(worker) {
+  return worker.lastExportedBlock + constants.BLOCK_INTERVAL < worker.lastConfirmedBlock;
+}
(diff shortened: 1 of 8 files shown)


blockchains/eth/lib/next_interval_calculator.js
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @param {*} worker A worker object, member variables would be modified
+ * @returns An object like so:
+ * {
+ *  success: Boolean,
+ *  fromBlock: Integer,
+ *  toBlock: Integer
+ * }
+ */
+async function nextIntervalCalculator(worker) {
```
YordanPavlov, line 15: I am externally modifying the member variables of a class here. Not sure if this is a good pattern but could not think of another way to reuse code in our current design. Ideas are welcome.

spiderjako (approved):

spiderjako (approved):

PR #136, opened 2023-06-12
Tweaking block number for UTXO

spiderjako:
Adding a '+1' to the scraping process, because we've made the START_BLOCK to be -1 at the beginning and the utxo worker would make an initial array of [-1,0,1...], whilst -1 is an invalid block number


spiderjako:
Fixed in another [PR](https://github.com/santiment/san-chain-exporter/pull/137)

PR #138, opened 2023-06-29
Implement a retry logic in the XRP exporter

YordanPavlov:
Sometimes the XRP endpoint would return an empty block. We are able to identify this by a non-0 `transaction_hash` field. In such case we can retry.

### DIFF
--- a/blockchains/xrp/xrp_worker.js
+++ b/blockchains/xrp/xrp_worker.js
@@ -10,7 +10,6 @@ class XRPWorker extends BaseWorker {
     super();
     this.nodeURLs = constants.XRP_NODE_URLS.split(',');
     this.connections = [];
-    this.emptyBlock = { ledger: { transactions: [] }};
     this.emptyTransactionHash = '0000000000000000000000000000000000000000000000000000000000000000';
     this.retryIntervalMs = 1000;
   }
--- a/blockchains/xrp/xrp_worker.js
+++ b/blockchains/xrp/xrp_worker.js
@@ -55,12 +55,14 @@ class XRPWorker extends BaseWorker {
   }
 
   async fetchLedger(connection, ledger_index, should_expand) {
-    const ledger = await this.connectionSend(connection, {
+    const result = await this.connectionSend(connection, {
       command: 'ledger',
       ledger_index: parseInt(ledger_index),
       transactions: true,
       expand: should_expand
-    }).then(value => value.result.ledger);
+    });
+
+    const ledger = result.result.ledger;
 
     assert(ledger.closed === true);
     assert(typeof ledger.transactions !== 'undefined');
--- a/blockchains/xrp/xrp_worker.js
+++ b/blockchains/xrp/xrp_worker.js
@@ -10,7 +10,6 @@ class XRPWorker extends BaseWorker {
     super();
     this.nodeURLs = constants.XRP_NODE_URLS.split(',');
     this.connections = [];
-    this.emptyTransactionHash = '0000000000000000000000000000000000000000000000000000000000000000';
     this.retryIntervalMs = 1000;
   }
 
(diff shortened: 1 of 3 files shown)


blockchains/xrp/xrp_worker.js
```diff
@@ -51,23 +53,33 @@ class XRPWorker extends BaseWorker {
...
-      return { ledger: ledger, transactions: [] };
+    assert(ledger.closed === true);
+    assert(typeof ledger.transactions !== 'undefined');
+    assert(typeof ledger.transaction_hash !== 'undefined');
+
+    if (ledger.transactions.length === 0 && ledger.transaction_hash !== this.emptyTransactionHash) {
+      // This block is invalid, the problem must be in the Endpoint. Wait and retry.
+      await new Promise((resolve) => setTimeout(resolve, this.retryIntervalMs));
+      logger.info(`Ledger ${ledger_index} is being retried due to invalid response`);
+      return await this.fetchLedger(connection, ledger_index, should_expand);
```
spiderjako, line 74: I know that in theory, k8s would kill the pod if it takes too much time, but isn't it a bad idea to leave an open recursion like this? Wouldn't a function with a retry mechanism be better? If it fails X times, then adios?

blockchains/xrp/xrp_worker.js
```diff
@@ -51,23 +53,33 @@ class XRPWorker extends BaseWorker {
...
-    }).then(value => value.result.ledger);
+      expand: should_expand
+    });
 
-    assert(ledger.closed === true);
+    const ledger = result.result.ledger;
 
-    if (typeof ledger.transactions === 'undefined' || ledger.transactions.length === 0) {
-      return { ledger: ledger, transactions: [] };
+    assert(ledger.closed === true);
```
spiderjako, line 66: Isn't using assertions in non-testing code discouraged? It's not really elegant in error-handling terms

test/xrp/worker.test.js
```diff
@@ -79,4 +79,52 @@ describe('workLoopSimpleTest', function () {
...
+    let sendCallsCount = 0;
+    // Reduce the retry interval so that tests finishes fast
+    worker.retryIntervalMs = 100;
+    worker.connectionSend = () => {
+      sendCallsCount += 1;
+      return Promise.resolve(invalidBlock);
+    };
+
+    let sendCallsCountWhileInvalid = 0;
+    // After a timeout, switch the mock function to return the vadlic block. Remember how many calls were made up until that moment.
```
spiderjako, line 114: valid*, I guess?

blockchains/xrp/xrp_worker.js
WonderBeat: retry looks better, agree

blockchains/xrp/xrp_worker.js
```diff
@@ -82,24 +94,24 @@ class XRPWorker extends BaseWorker {
...
-    }).then(value => value.result.ledger);
+      return resolvedTransactions.map(t => t.result);
+  }
 
-    assert(result.closed === true);
+  async fetchLedgerTransactions(connection, ledger_index) {
+    const ledger = await this.fetchLedger(connection, ledger_index, false);
 
-    return { ledger: ledger, transactions: result.transactions };
+    if (ledger.transactions.length > 200) {
```
WonderBeat, line 106: is it better to batch each 200 and process the same way?

blockchains/xrp/xrp_worker.js
```diff
@@ -10,6 +10,8 @@ class XRPWorker extends BaseWorker {
     super();
     this.nodeURLs = constants.XRP_NODE_URLS.split(',');
     this.connections = [];
+    this.emptyTransactionHash = '0000000000000000000000000000000000000000000000000000000000000000';
```
WonderBeat, line 13: i'd prefer something like '0'.repeat(160) or something like this. Let PC do hard work for you ;)

PR #139, opened 2023-06-30
Capture SIGTERM signal coming from Kubernetes

YordanPavlov:
Up until now the container entrypoint looked like so:
```npm -> micro -> index.js```
this prevented the `SIGTERM` coming from Kubernetes to reach our `index.js` and allow for graceful stop. After reading on the topic I could not find a way how to propagate the signal besides changing the entry point to just:
```node index.js```

In this PR the above is implemented, this requires for the Micro service to be started and stopped explicitly. Also some refactoring to simplify the main program flow, now we `await` various method calls and know where exactly execution is, allowing to print 'Bye' at the user at the end to confirm exit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,7 @@
   "main": "index.js",
   "scripts": {
     "test": "CONTRACT_MAPPING_FILE_PATH=test/erc20/contract_mapping/contract_mapping.json LOG_LEVEL=error CONTRACT_MODE=extract_exact_overwrite mocha --recursive --reporter spec",
-    "lint": "eslint '**/*.js'",
-    "start": "micro"
+    "lint": "eslint '**/*.js'"
   },
   "repository": {
     "type": "git",
--- a/bin/export_env_vars.sh
+++ b/bin/export_env_vars.sh
@@ -1,6 +1,6 @@
-export KAFKA_URL=localhost:9092
-export ZOOKEEPER_URL=localhost:2181
-export NODE_URL=http://erigon-hz.stage.san:30250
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -5,8 +5,7 @@
   "main": "index.js",
   "scripts": {
     "test": "CONTRACT_MAPPING_FILE_PATH=test/erc20/contract_mapping/contract_mapping.json LOG_LEVEL=error CONTRACT_MODE=extract_exact_overwrite mocha --recursive --reporter spec",
-    "lint": "eslint '**/*.js'",
-    "start": "micro"
```
spiderjako, line 9: Isn't it better to still have the "start" script, just by using ```node index.js```? Or do you want to escape NPM in addition to micro?

YordanPavlov: This wouldn't work, at least without further tweaks, look at how processes look:

``
 PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                                                                                                                                          
     20 root      20   0   11.5g  69764   6404 S   0.3   0.1  28:54.91 node                                                                                                                                                                             
      1 root      20   0    2400     80      0 S   0.0   0.0   0:00.00 sh                                                                                                                                                                               
      8 root      20   0  786312  30792      0 S   0.0   0.0   0:00.41 npm start 
      ```
      
Kubernetes would send the signal to the process with PID 1, this needs to be either `sh` (which would transfer it), or `node`

spiderjako (approved):

PR #141, opened 2023-07-10
Prevent Cardano from writing invalid lastExporedBlock

YordanPavlov:
When the Cardano node is out of sync it would report past last block. This would cause the Cardano exporter to abandon its progress and resume from beginning (block 0). By looking at the code I couldn't identify under what situation this happens, we already added some guards against this. However I am removing one potential problem line.

### DIFF
--- a/blockchains/cardano/cardano_worker.js
+++ b/blockchains/cardano/cardano_worker.js
@@ -153,8 +153,12 @@ class CardanoWorker extends BaseWorker {
     return response.data.transactions;
   }
 
-  async init() {
+  async setLastConfirmedBlock() {
     this.lastConfirmedBlock = await this.getCurrentBlock() - constants.CONFIRMATIONS;
+  }
+
+  async init() {
+    await this.setLastConfirmedBlock();
   }
 
   async work() {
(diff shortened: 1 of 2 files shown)


blockchains/cardano/cardano_worker.js
```diff
@@ -168,12 +172,11 @@ class CardanoWorker extends BaseWorker {
...
       // API would return a partial last block. The same would not happen if the same block is fetched as part of 2
       // block interval.
-      const newConfirmedBlock = await this.getCurrentBlock() - constants.CONFIRMATIONS;
-      if (newConfirmedBlock < fromBlock + 2) {
+      await this.setLastConfirmedBlock();
+      if (this.lastConfirmedBlock < fromBlock + 2) {
         // The Node has not progressed enough
         return [];
       }
-      this.lastConfirmedBlock = newConfirmedBlock;
```
YordanPavlov, line 176: No need to go through new variable here, the `this.lastConfirmedBlock` would already be polluted on init() if the Node is broken.

blockchains/cardano/cardano_worker.js
```diff
@@ -191,7 +194,6 @@ class CardanoWorker extends BaseWorker {
     else {
       transactions = await this.getTransactions(fromBlock, this.lastConfirmedBlock);
       if (transactions.length === 0) {
-        this.lastExportedBlock = this.lastConfirmedBlock;
```
YordanPavlov, line 194: That's one potential place where can break the `this.lastExporedBlock` logic. However we shouldn't reach here if `this.lastConfirmedBlock < this.lastExportedBlock`. Anyways removing this as it is unsafe.

spiderjako (approved):

antisaa (approved):

PR #151, opened 2023-09-17
Adding S3 fetch functionality

spiderjako:


lib/fetch_blocks_list.js
```diff
@@ -0,0 +1,239 @@
...
+        throw err;
+      });
+
+    const dataKeys = Object.keys(data);
+    let contents, isTruncated;
+    if (dataKeys.indexOf('Contents') >= 0 && dataKeys.indexOf('IsTruncated') >= 0) {
+      contents = data.Contents;
+      isTruncated = data.IsTruncated;
+    } else {
+      logger.error(`Error fetching S3 data: Missing key in response: ${JSON.stringify(data)}`);
```
WonderBeat, line 55: image

```
function a() { 
  console.error("Error message");
  throw Error("Error message");             
}
a();
```

This style of logging leads to error message duplicates. In general we should never log before we throw exception. The reason is simple. Most (all) code that catch our exceptions will log stacktrace and error msg upon the catch. Even if we don't catch it ourself, vm will do it for us and print the stacktrace and error message.
I made a small example to show the issue

lib/fetch_blocks_list.js
```diff
@@ -0,0 +1,239 @@
...
+    const result = {};
+    for (const key of keys) {
+      const currObject = await this.s3Client.send(new GetObjectCommand({
+        Bucket: S3_BUCKET,
+        Key: key
+      }));
+      const csvData = await this.processDataIntoArray(currObject, key);
+      result[key] = csvData
+        .sort((a, b) => a - b)
+        .map(x => parseInt(x));
```
WonderBeat, line 136:
```
parseInt("a")
NaN 
```
but I'm mostly sure we want to fail if we have such data in the stream

lib/fetch_blocks_list.js
```diff
@@ -0,0 +1,239 @@
...
+    }
+    return result;
+  }
+
+  /**
+   * @param {number} start Start block number
```
WonderBeat, line 142: inclusive or exclusive?
image

lib/fetch_blocks_list.js
```diff
@@ -0,0 +1,239 @@
...
+   */
+  makeIntervals(data) {
+    if (data.length === 0) {
+      return [];
+    }
+    const intervals = [];
+    let start = data[0];
+    let end = data[0];
+    for (let i = 1; i < data.length; i++) {
+      if (data[i] - data[i - 1] === 1) {
```
WonderBeat, line 177: can we assert that `data[i] > data[i-1]`?

WonderBeat (approved):

lib/fetch_blocks_list.js
spiderjako: You're right, for this PR I'll remove the redundant logs only from the new code, this is an idea for another PR to remove all such redundant loggings

lib/fetch_blocks_list.js
spiderjako: Yes, the array should be sorted in ascending order beforehand
```js
---processAllS3Data---
result[key] = csvData
  .sort((a, b) => a - b)
  .map(x => parseInt(x));
```

blockchains/eth/lib/constants.js
```diff
@@ -1,16 +1,13 @@
...
 const BURN_ADDRESS = 'burn';
-const ETH_WITHDRAWAL = 'withdrawal'
-const IS_ETH = parseInt(process.env.IS_ETH || '1');
+const ETH_WITHDRAWAL = 'withdrawal';
 const LONDON_FORK_BLOCK = 12965000;
 const SHANGHAI_FORK_BLOCK = 17034871;
-// This is the API method that Parity provides for fetching receipts across multiple blocks. Erigon instead provides
-// a method called 'eth_getBlockReceipts'. If we are deploying against Erigon, we need to overwrite this variable
-// through the deploy.
+const IS_ETH = parseInt(process.env.IS_ETH || '1');
```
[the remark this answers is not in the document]
spiderjako: About this I'm not really sure, this was done a long time ago I think, I've done `EXPORT_BLOCKS_LIST` in the same manner, but wasn't really thinking about it. I guess it could just be a boolean, will try it out

lib/fetch_blocks_list.js
```diff
@@ -0,0 +1,239 @@
...
+    this.blocksList = [];
+  }
+
+  /**
+   * @returns {void} Initializes the S3 client and its parameters
+   */
+  initS3Client() {
+    this.s3Client = new S3Client({
+      credentials: {
+        accessKeyId: AWS_ACCESS_KEY_ID,
```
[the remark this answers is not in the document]
spiderjako: You're partially right, but what would the alternative be? The S3Client would be initialised and won't throw an error, even if it's not in an object.

blockchains/erc20/erc20_worker.js
```diff
@@ -19,15 +20,24 @@ class ERC20Worker extends BaseWorker {
...
     setGlobalTimestampManager(exporter);
+    if (constants.EXPORT_BLOCKS_LIST === '1') {
+      await this.blocksListExporter.initInWorker();
+    }
   }
 
   async work() {
-    const result = await nextIntervalCalculator(this);
+    const result = (constants.EXPORT_BLOCKS_LIST === '1') ?
+      this.blocksListExporter.calculateArrayPosition() :
```
YordanPavlov, line 38: After 1+ h in this PR I start to realize that this feature is meant to read the block numbers to extract ONLY from S3 and the data is still read from the Node, what's the purpose of this? That block list could easily have been provided through the deploy?

spiderjako: You can't really give it a list of 5mil blocks through the deploy though. I've initially thought the same, but how would you go onto specifying so much data in the deploy?

YordanPavlov: What's the use case here? Give 5mil random blocks? You always extract a range.

spiderjako: We need a way to extract just a portion of the blocks, not all of them. In the case of a BEP20 rerun, you'll need to extract info for 30mil blocks, when sometimes all you need is just 5mil of them. Sometimes, it's not a range but rather they're spread out. Or maybe I don't get the question?

YordanPavlov: > you'll need to extract info for 30mil blocks, when sometimes all you need is just 5mil of them.

You can easily provide the range start and range end through the deploy. If you have 10+ ranges, you may choose to use a ConfigMap for readability.

> it's not a range but rather they're spread out

I can't imagine how this has happened, every 3rd block is wrong or what? Can't you break this up into multiple ranges?

YordanPavlov: If you have a huge number of randomly broken blocks, something is generally wrong with the exporter. I don't this this is a use case we want to have as part of our playbook.

spiderjako: > If you have a huge number of randomly broken blocks, something is generally wrong with the exporter.
As is the current case, you don't really know that until it happens. And sometimes you don't have the resources to get a full range of blocks. I'm not sure why this wouldn't be beneficial, as it minimises the data that's going to be extracted
>  If you have 10+ ranges, you may choose to use a ConfigMap for readability.
I'm not sure how you'd do that with 5mil blocks, if there are a lot of gaps and you'd have to manually consider these.

YordanPavlov: > As is the current case, you don't really know that until it happens. 

What does that mean? If you have an exporter that randomly fails blocks, that's pretty bad, just re-export everything. If you are about to say that this is slow - you may have worked on implementing parallel export, this would have been much better use of time. We own the Full Node, we may hit it hard.

> And sometimes you don't have the resources to get a full range of blocks

How come?

> I'm not sure why this wouldn't be beneficial, as it minimises the data that's going to be extracted

I'm not against specifying a range of blocks, this makes tons of sense. The problem is why do you use a storage provider service (S3) to specify those blocks?

> I'm not sure how you'd do that with 5mil blocks, if there are a lot of gaps and you'd have to manually consider these.

Reexport everything.

blockchains/erc20/erc20_worker.js
```diff
@@ -24,10 +25,36 @@ class ERC20Worker extends BaseWorker {
   async init(exporter) {
     this.lastConfirmedBlock = await this.web3.eth.getBlockNumber() - constants.CONFIRMATIONS;
     setGlobalTimestampManager(exporter);
+
+    if (constants.EXPORT_BLOCKS_LIST) {
+      this.blocksList = await initBlocksList().catch((err) => { throw err; });
```
WonderBeat, line 30: `.catch((err) => { throw err; });`
does this make sense? as I read this - we catch to rethrow the same exception

blockchains/erc20/erc20_worker.js
```diff
@@ -24,10 +25,36 @@ class ERC20Worker extends BaseWorker {
...
+    if (constants.EXPORT_BLOCKS_LIST) {
+      this.blocksListPosition = initBlocksListPosition(lastProcessedPosition, this.blocksList);
+      this.lastExportedBlock = this.blocksList[this.blocksListPosition][0];
+      this.lastPrimaryKey = parseInt(process.env.START_PRIMARY_KEY || '-1');
+    }
+    super.initPosition(lastProcessedPosition);
+  }
+
+  checkBlocksListPosition() {
+    return this.blocksListPosition < this.blocksList.length ?
```
WonderBeat, line 44: imo it looks better with `if/else`

lib/fetch_blocks_list.js
```diff
@@ -0,0 +1,108 @@
...
+      .pipe(csv())
+      .on('data', (data) => {
+        const blockNumber = Object.values(data)[0];
+        csvData.push(blockNumber);
+      })
+      .on('end', () => {
+        logger.info('Processed data from config file');
+        resolve(csvData);
+      })
+      .on('error', (err) => { console.log(err); reject(err); });
```
WonderBeat, line 25: are we sure we want to log it this place? I mean that once again I expect that if we throw exception then it will be logged at the place you catch it. in this case we'll end up logging it twice

lib/fetch_blocks_list.js
```diff
@@ -0,0 +1,108 @@
...
+  });
+};
+
+/**
+ * @param {number} start Start block number
+ * @param {number} end End block number
+ * @param {number} interval Maximum distance between blocks to be fetched
+ * @returns {Array} Array of intervals
+ */
+const breakdownInterval = (start, end) => {
```
WonderBeat, line 35: what's the reason to declare functions like this?

WonderBeat (approved):

lib/fetch_blocks_list.js
spiderjako: Yes, I've missed removing it

blockchains/erc20/erc20_worker.js
```diff
@@ -24,10 +25,37 @@ class ERC20Worker extends BaseWorker {
...
+    }
+  }
+
+  async initPosition(lastProcessedPosition) {
+    if (constants.EXPORT_BLOCKS_LIST) {
+      this.blocksListPosition = initBlocksListPosition(lastProcessedPosition, this.blocksList);
+      this.lastExportedBlock = this.blocksList[this.blocksListPosition][0];
+      this.lastPrimaryKey = parseInt(process.env.START_PRIMARY_KEY || '-1');
+    }
+    super.initPosition(lastProcessedPosition);
```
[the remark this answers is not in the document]
spiderjako: My mistake, it should be in an `else` clause.

blockchains/erc20/erc20_worker.js
```diff
@@ -24,17 +25,45 @@ class ERC20Worker extends BaseWorker {
   async init(exporter) {
     this.lastConfirmedBlock = await this.web3.eth.getBlockNumber() - constants.CONFIRMATIONS;
     setGlobalTimestampManager(exporter);
+
+    if (constants.EXPORT_BLOCKS_LIST) {
+      this.blocksList = await initBlocksList();
+    }
+  }
+
+  getBlocksListInterval() {
```
YordanPavlov, line 34: Minor. I would have made this a stand alone function operating on two arguments. This makes reuse in other classes easier.

blockchains/erc20/erc20_worker.js
```diff
@@ -53,6 +82,7 @@ class ERC20Worker extends BaseWorker {
       this.lastPrimaryKey = events[events.length - 1].primaryKey;
     }
 
+    if (constants.EXPORT_BLOCKS_LIST) this.blocksList.shift();
```
YordanPavlov, line 85: Why the extra shift, you have this logic in the `getBlocksListInterval`

YordanPavlov (approved):

PR #153, opened 2023-10-31
Simplify timestamps caching

YordanPavlov:
In this PR:
* Remove a complex logic inside `TimestampsManager` that we do not use. It was indented to work around Node timestamps problems, but was never deployed. Instead rework this class to be a simple timestamps cache.
* Use the newly available cache instead of caching inside a plain `{}` object.


blockchains/erc20/erc20_worker.js
```diff
@@ -21,9 +22,8 @@ class ERC20Worker extends BaseWorker {
     this.web3 = new Web3(new Web3.providers.HttpProvider(constants.NODE_URL));
   }
 
-  async init(exporter) {
+  async init() {
     this.lastConfirmedBlock = await this.web3.eth.getBlockNumber() - constants.CONFIRMATIONS;
-    setGlobalTimestampManager(exporter);
```
YordanPavlov, line 26: Remove the notion of a 'global timestamps manager', it was never truly used.

blockchains/erc20/erc20_worker.js
```diff
@@ -37,11 +37,12 @@ class ERC20Worker extends BaseWorker {
     let events = [];
     let overwritten_events = [];
     if ('extract_exact_overwrite' === constants.CONTRACT_MODE) {
-      events = await contractEditor.getPastEventsExactContracts(this.web3, result.fromBlock, result.toBlock);
+      events = await contractEditor.getPastEventsExactContracts(this.web3, result.fromBlock, result.toBlock,
+        new TimestampsCache());
```
YordanPavlov, line 41: Provide a cache which is intended to live just for the current set of events. The next set of events would be for new blocks so no need to persist it.

blockchains/erc20/lib/fetch_events.js
```diff
@@ -12,33 +12,9 @@ const BNB_contract = '0xb8c77482e45f1f44de1745f52c74426c631bdd52';
...
-  }
-  else {
-    return await timestampsManager.getTimestampFromNode(web3, blockNumber);
-  }
-
-}
-
-async function decodeEventBasicInfo(web3, event, blockTimestamps) {
-  let timestamp;
-  if (!blockTimestamps[event['blockNumber']]) {
```
YordanPavlov, line 37: Up until now, actually this `blockTimestamps` argument was the cache, and it was a simple `{}`. IMO the new argument better denotes its purpose.

blockchains/erc20/lib/fetch_events.js
```diff
@@ -228,15 +204,14 @@ async function getRawEvents(web3, fromBlock, toBlock, contractAddress) {
   return await web3.eth.getPastLogs(queryObject);
 }
 
-async function decodeEvents(web3, events, decodeFunctions=decodeFunctionsMap) {
-  const blockTimestamps = {};
```
YordanPavlov, line 232: Previously that object was the cache. However that was not optimal for the case where we extract per contract, because we arrive in this method for each contract. This invalidates the cache when it can be still used.

blockchains/erc20/lib/timestamps_cache.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+
+  async getBlockTimestamp(web3, blockNumber) {
+    const timestampStore = this.getTimestampFromStore(blockNumber);
+    if (timestampStore !== DATA_MISSING) {
+      return timestampStore;
+    }
+
+    let timestamp = await this.getTimestampFromNode(web3, blockNumber);
```
spiderjako, line 34: Shouldn't this be 'const', as far as I can see, its value isn't changed later

blockchains/erc20/lib/timestamps_cache.js
```diff
@@ -0,0 +1,45 @@
...
+  getTimestampFromStore(blockNumber) {
+    if (Object.prototype.hasOwnProperty.call(this.timestampStore, blockNumber)) {
+      return this.timestampStore[blockNumber];
+    }
+
+    return DATA_MISSING;
+  }
+
+  async getTimestampFromNode(web3, blockNumber) {
+    console.log('Getting timestamp for ', blockNumber);
```
spiderjako, line 22: Shouldn't we use the already defined `logger`

YordanPavlov: I used it for debugging, removed now.

spiderjako (approved):

PR #161, opened 2023-11-14
Exact mode no overwrite

YordanPavlov:
The main goal of this PR is to extend the functionality of the exporter and make it suitable to extract specific contracts. We have something similar already but it's a bit different. What we have is extract a contract and change its address to something else, we use this to map multiple contracts to a single token name. The functionality I add now would again extract events for only specific contracts, but do not modify the contract address.

In this PR:
* Attempt to simplify the logic around the contract overwriting.
* Add new extraction mode - where we extract specific contracts but do not overwrite them.
* Write all messages for a contract in the same partition

### DIFF
--- a/blockchains/erc20/erc20_worker.js
+++ b/blockchains/erc20/erc20_worker.js
@@ -19,6 +19,8 @@ class ERC20Worker extends BaseWorker {
 
     logger.info(`Connecting to Ethereum node ${constants.NODE_URL}`);
     this.web3 = new Web3(new Web3.providers.HttpProvider(constants.NODE_URL));
+    this.contractsOverwriteArray = [];
+    this.contractsUnmodified = [];
   }
 
   async init() {
@@ -31,11 +33,15 @@ class ERC20Worker extends BaseWorker {
     if (constants.CONTRACT_MODE !== 'vanilla') {
       const parsedContracts = await readJsonFile(constants.CONTRACT_MAPPING_FILE_PATH);
 
-      this.contractsOverwriteArray = parsedContracts.modified_contracts.map((contract) => new ContractOverwrite(contract));
-      this.contractsUnmodified = parsedContracts.unmodified_contracts.map((contract) => contract.toLowerCase());
+      if (parsedContracts.modified_contracts) {
+        this.contractsOverwriteArray = parsedContracts.modified_contracts.map((contract) => new ContractOverwrite(contract));
+      }
+      if (parsedContracts.unmodified_contracts) {
+        this.contractsUnmodified = parsedContracts.unmodified_contracts.map((contract) => contract.toLowerCase());
+      }
 
       logger.info(`Running in '${constants.CONTRACT_MODE}' contracts mode', ` +
-        `${this.contractsOverwriteArray.length} contracts will be monitored.`);
+        `${this.contractsOverwriteArray.length + this.contractsUnmodified.length} contracts will be monitored.`);
       logger.info(`Overwritten contracts are: ${JSON.stringify(this.contractsOverwriteArray)}`);
     }
   }
--- a/blockchains/erc20/erc20_worker.js
+++ b/blockchains/erc20/erc20_worker.js
@@ -31,7 +31,8 @@ class ERC20Worker extends BaseWorker {
     if (constants.CONTRACT_MODE !== 'vanilla') {
       const parsedContracts = await readJsonFile(constants.CONTRACT_MAPPING_FILE_PATH);
 
-      this.contractsOverwriteArray = parsedContracts.map((parsedContract) => new ContractOverwrite(parsedContract));
+      this.contractsOverwriteArray = parsedContracts.modified_contracts.map((contract) => new ContractOverwrite(contract));
+      this.contractsUnmodified = parsedContracts.unmodified_contracts.map((contract) => contract.toLowerCase());
 
       logger.info(`Running in '${constants.CONTRACT_MODE}' contracts mode', ` +
         `${this.contractsOverwriteArray.length} contracts will be monitored.`);
@@ -82,6 +83,11 @@ class ERC20Worker extends BaseWorker {
           editAddressAndAmount(event, contractOverwrite);
           events.push(...rawEvents);
         }
+      }
+      for (const contractUnmodified of this.contractsUnmodified) {
+        const rawEvents = await getPastEvents(this.web3, result.fromBlock, result.toBlock, contractUnmodified,
+          timestampsCache);
+        events.push(...rawEvents);
       }
     }
     else {
--- a/blockchains/erc20/erc20_worker.js
+++ b/blockchains/erc20/erc20_worker.js
@@ -11,6 +11,11 @@ const { TimestampsCache } = require('./lib/timestamps_cache');
 const { getPastEvents } = require('./lib/fetch_events');
 const { initBlocksList } = require('../../lib/fetch_blocks_list');
 
+/**
+ * A simple non cryptographic hash function similar to Java's 'hashCode'
+ * @param input A string input
+ * @returns A 32 bit positive integer
+ */
 function simpleHash(input) {
   var hash = 0, i, chr;
 
--- a/blockchains/erc20/erc20_worker.js
+++ b/blockchains/erc20/erc20_worker.js
@@ -57,7 +57,7 @@ class ERC20Worker extends BaseWorker {
         this.contractsUnmodified = parsedContracts.unmodified_contracts.map((contract) => contract.toLowerCase());
       }
 
-      exporter.hashFunction = (event) => simpleHash(event.contract);
+      await exporter.initPartitioner((event) => simpleHash(event.contract));
 
       logger.info(`Running in '${constants.CONTRACT_MODE}' contracts mode', ` +
         `${this.contractsOverwriteArray.length + this.contractsUnmodified.length} contracts will be monitored.`);
--- a/blockchains/erc20/erc20_worker.js
+++ b/blockchains/erc20/erc20_worker.js
@@ -34,6 +34,7 @@ class ERC20Worker extends BaseWorker {
     super();
 
     logger.info(`Connecting to Ethereum node ${constants.NODE_URL}`);
+    logger.info(`Applying the following settings: ${JSON.stringify(constants)}`);
     this.web3 = new Web3(new Web3.providers.HttpProvider(constants.NODE_URL));
     this.contractsOverwriteArray = [];
     this.contractsUnmodified = [];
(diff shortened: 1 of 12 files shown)


blockchains/erc20/lib/contract_overwrite.js
```diff
@@ -23,87 +19,50 @@ class ContractOverwrite {
   }
 }
 
-class ContractEditor {
```
YordanPavlov, line 26: Removing this class and using its methods as functions. I think the code is subjectively easier to understand now.

blockchains/erc20/erc20_worker.js
```diff
@@ -63,17 +78,28 @@ class ERC20Worker extends BaseWorker {
...
 
-    let events;
+    let events = [];
     let overwritten_events = [];
     if ('extract_exact_overwrite' === constants.CONTRACT_MODE) {
-      events = await contractEditor.getPastEventsExactContracts(this.web3, result.fromBlock, result.toBlock,
-        new TimestampsCache());
-      contractEditor.changeContractAddresses(events);
+      const timestampsCache = new TimestampsCache();
+      for (const contractOverwrite of this.contractsOverwriteArray) {
```
YordanPavlov, line 85: I am changing the logic of iteration a bit here. Before we called `contractEditor.getPastEventsExactContracts` and getting the events for all the contracts. We did not bulk the calls (see removed code), instead iterated per contract. However the information for which events are for which contract was lost, then we had to do some logic to map events to contract so to overwrite. This is not effective and somewhat harder to comprehend.

blockchains/erc20/erc20_worker.js
```diff
@@ -29,6 +29,21 @@ class ERC20Worker extends BaseWorker {
...
       this.blocksList = await initBlocksList();
     }
+    console.log(JSON.stringify(constants));
+    if (constants.CONTRACT_MODE !== 'vanilla') {
+      const parsedContracts = await readJsonFile(constants.CONTRACT_MAPPING_FILE_PATH);
+
+      if (parsedContracts.modified_contracts) {
+        this.contractsOverwriteArray = parsedContracts.modified_contracts.map((contract) => new ContractOverwrite(contract));
+      }
+      if (parsedContracts.unmodified_contracts) {
```
YordanPavlov, line 39: That's the new functionality, where we extract per contract, without overwriting.

lib/kafka_storage.js
```diff
@@ -65,13 +65,10 @@ class Exporter {
         retries: ZOOKEEPER_RETRIES
       }
     );
-  }
 
-  get topicName() {
```
YordanPavlov, line 70: Have it as a member variable instead of method

test/erc20/contract_mapping/contract_mapping.json
```diff
@@ -1,213 +1,217 @@
...
-        },
-        {
-            "address": "0x91af0fbb28aba7e31403cb457106ce79397fd4e6",
-            "multiplier": 1
-        }
-    ],
-    "new_address": "aergo_contract"
-}]
-
+    "unmodified_contracts": [],
```
YordanPavlov, line 2: We support providing an array of strings which would be contracts that we extract

blockchains/erc20/erc20_worker.js
```diff
@@ -63,17 +94,23 @@ class ERC20Worker extends BaseWorker {
...
 
-    let events;
+    let events = [];
     let overwritten_events = [];
     if ('extract_exact_overwrite' === constants.CONTRACT_MODE) {
-      events = await contractEditor.getPastEventsExactContracts(this.web3, result.fromBlock, result.toBlock,
-        new TimestampsCache());
-      contractEditor.changeContractAddresses(events);
+      const timestampsCache = new TimestampsCache();
+      events = await getPastEvents(this.web3, result.fromBlock, result.toBlock, this.allOldContracts, timestampsCache);
```
YordanPavlov, line 101: We fetch the events for all overwritten contracts at once now. Apparently the Node API supports it, so we can reduce the number of requests. Before this change we used to do one request per overwritten contract.

blockchains/erc20/erc20_worker.js
```diff
@@ -3,24 +3,40 @@ const Web3 = require('web3');
...
+const { stableSort, readJsonFile } = require('./lib/util');
 const BaseWorker = require('../../lib/worker_base');
 const { nextIntervalCalculator } = require('../eth/lib/next_interval_calculator');
 const { TimestampsCache } = require('./lib/timestamps_cache');
 const { getPastEvents } = require('./lib/fetch_events');
 const { initBlocksList } = require('../../lib/fetch_blocks_list');
 
+/**
+ * A simple non cryptographic hash function similar to Java's 'hashCode'
+ * @param input A string input
```
spiderjako, line 16: Should be something like
```js
* @param {string} input
* @returns {number} A 32 bit non-negative integer
```
also, maybe something more specific than "string input"? Could it be a random string or it has limitations?

blockchains/erc20/erc20_worker.js
```diff
@@ -3,24 +3,40 @@ const Web3 = require('web3');
...
+ * @returns A 32 bit positive integer
+ */
+function simpleHash(input) {
```
spiderjako, line 19: Maybe another name? `simpleHash` is kind of too broad, `bitHash` maybe? This comment is more of an opinion than a change request

blockchains/erc20/erc20_worker.js
```diff
@@ -63,17 +99,23 @@ class ERC20Worker extends BaseWorker {
...
     logger.info(`Fetching transfer events for interval ${result.fromBlock}:${result.toBlock}`);
 
-    let events;
+    let events = [];
     let overwritten_events = [];
     if ('extract_exact_overwrite' === constants.CONTRACT_MODE) {
-      events = await contractEditor.getPastEventsExactContracts(this.web3, result.fromBlock, result.toBlock,
-        new TimestampsCache());
-      contractEditor.changeContractAddresses(events);
+      const timestampsCache = new TimestampsCache();
```
spiderjako, line 105: Couldn't we just add this `new TimestampCache()` before the if-else statements and give it to the functions?

blockchains/erc20/lib/constants.js
```diff
@@ -14,7 +14,7 @@ const NODE_URL = process.env.NODE_URL || process.env.PARITY_URL || 'http://local
 
 const CONTRACT_MAPPING_FILE_PATH = (
     process.env.CONTRACT_MAPPING_FILE_PATH ?
-        '../../../' + process.env.CONTRACT_MAPPING_FILE_PATH :
```
spiderjako, line 17: This would require changes in the DevOps configmap part, right?

blockchains/erc20/lib/contract_overwrite.js
spiderjako: I thought about this myself a few days ago. nice

blockchains/erc20/lib/contract_overwrite.js
```diff
@@ -23,87 +19,67 @@ class ContractOverwrite {
...
-  changeContractAddresses(events) {
-    for (let event of events) {
-      for (const contractOverwrite of this.contractsOverwriteArray) {
-        if (this.isContractMatchesExactList(event, contractOverwrite)) {
-          this.editAddressAndAmount(event, contractOverwrite);
-          break;
-        }
+/**
+ *
+ * @param events A list of events to go over and check contract address. Events needing contract change will
```
spiderjako, line 24: ditto

blockchains/erc20/lib/contract_overwrite.js
```diff
@@ -23,87 +19,67 @@ class ContractOverwrite {
...
+}
 
+function editAddressAndAmount(event, contractOverwrite) {
+  const multiplier = contractOverwrite.mapAddressToMultiplier[event.contract];
+  if (!multiplier) {
+    throw new Error('Event contract does not match expected');
+  }
+  event.contract = contractOverwrite.newContract;
   /**
+   * Note 1: Whether we should round here is up for discussion. The amounts in our pipeline are 'float' anyways but up until this feature
```
spiderjako, line 45: Shouldn't this be a ticket in itself? Don't think this belongs in the code. Maybe leave a comment that it could be changed or how it works now, but 5 lines of notes is overwhelming for a code file IMO

blockchains/erc20/lib/contract_overwrite.js
```diff
@@ -23,87 +19,67 @@ class ContractOverwrite {
...
-    for (let event of events) {
-      for (const contractOverwrite of this.contractsOverwriteArray) {
-        if (this.isContractMatchesExactList(event, contractOverwrite)) {
-          event = JSON.parse(JSON.stringify(event));
-          this.editAddressAndAmount(event, contractOverwrite);
-          result.push(event);
-          break;
-        }
+/**
+ *
```
spiderjako, line 58: not needed I think. Ditto for next lines

blockchains/erc20/lib/contract_overwrite.js
```diff
@@ -23,87 +19,67 @@ class ContractOverwrite {
...
+ * @param events A list of events to go over and check contract address.
+ * @returns A deep copy of the events which have had change applied.
+ */
+function extractChangedContractAddresses(events, contractsOverwriteArray) {
+  const result = [];
+
+  for (let event of events) {
+    for (const contractOverwrite of contractsOverwriteArray) {
+      if (contractOverwrite.oldAddresses.includes(event.contract)) {
+        event = JSON.parse(JSON.stringify(event));
```
spiderjako, line 68: I see that this has been in the class as well. Why do we stringify and then parse?

blockchains/erc20/lib/util.js
```diff
@@ -18,7 +20,18 @@ function stableSort(array, sortFunc) {
   array.forEach(x => delete x._position);
 }
 
+async function readJsonFile(filePath) {
+  try {
+    const data = await fs.readFile(filePath, 'utf8');
+    const parsedContracts = JSON.parse(data);
+    return parsedContracts;
+  } catch (err) {
+    console.error(err);
```
spiderjako, line 29: Why do we not throw something? Also, it'd be better if we sticked with logger I think

WonderBeat (approved):

lib/kafka_storage.js
```diff
@@ -241,13 +238,23 @@ class Exporter {
   }
 
   async sendDataWithKey(events, keyField) {
+    // We delay the finding of the partition count so that we are sure that we have a connected producer
+    if (this.partitionFunction && !this.partitionCount) {
+      this.partitionCount = await this.findPartitionCount(this.topicName);
```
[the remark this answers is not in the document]
YordanPavlov: nvm, I've added the Partitioner as suggested

blockchains/erc20/lib/contract_overwrite.js
spiderjako: I don't necessarily dislike 5 lines of comments, but they're in the middle of the function. But I do agree, that we need to have more serious docs

blockchains/erc20/lib/contract_overwrite.js
spiderjako: I think it makes sense, yes. Was just curious

spiderjako (approved):

blockchains/erc20/erc20_worker.js
WonderBeat: Thanks, I believe it's better add another dependency to keep our codebase small. but I don't mind to do it this way. thanks for the link

WonderBeat (approved):

antisaa (approved):

PR #162, opened 2023-11-18
Adding parallelism to the ETH pipeline

spiderjako:


blockchains/eth/eth_worker.js
```diff
@@ -29,6 +30,32 @@ class ETHWorker extends BaseWorker {
     this.withdrawalsDecoder = new WithdrawalsDecoder(this.web3, this.web3Wrapper);
   }
 
+  async ethClientRequestWithRetry(...params) {
+    let retries = 0;
+    let retryIntervalMs = 0;
+    while (retries < constants.MAX_RETRIES) {
+      try {
+        const response = await this.ethClient.request(...params);
```
YordanPavlov, line 38: should (...params) be spread again, I thought that request would work with an array

blockchains/eth/eth_worker.js
```diff
@@ -29,6 +30,32 @@ class ETHWorker extends BaseWorker {
...
+        if (response.error || response.result === null) {
+          retries++;
+          retryIntervalMs += (2000 * retries);
+          logger.error(`${params[0]} failed. Reason: ${response.error}. Retrying for ${retries} time`);
+          await new Promise((resolve) => setTimeout(resolve, retryIntervalMs));
+          continue;
```
YordanPavlov, line 44: Wouldn't `if-else`, be cleaner than `continue`?

blockchains/eth/eth_worker.js
```diff
@@ -29,6 +30,32 @@ class ETHWorker extends BaseWorker {
...
 
+  async ethClientRequestWithRetry(...params) {
+    let retries = 0;
+    let retryIntervalMs = 0;
+    while (retries < constants.MAX_RETRIES) {
+      try {
+        const response = await this.ethClient.request(...params);
+        if (response.error || response.result === null) {
+          retries++;
+          retryIntervalMs += (2000 * retries);
```
YordanPavlov, line 41: Use const value for 2000, like `constants.BACKOFF_RETRY_STEP`

blockchains/eth/eth_worker.js
```diff
@@ -29,6 +30,32 @@ class ETHWorker extends BaseWorker {
...
+          logger.error(`${params[0]} failed. Reason: ${response.error}. Retrying for ${retries} time`);
```
YordanPavlov, line 42: Is it really the `params[0]` that failed? That's one request out of the batch, maybe just summarize that a batch of size `params.length()` has failed.

blockchains/eth/eth_worker.js
```diff
@@ -29,6 +30,32 @@ class ETHWorker extends BaseWorker {
...
   }
 
+  async ethClientRequestWithRetry(...params) {
+    let retries = 0;
+    let retryIntervalMs = 0;
+    while (retries < constants.MAX_RETRIES) {
+      try {
+        const response = await this.ethClient.request(...params);
+        if (response.error || response.result === null) {
+          retries++;
```
YordanPavlov, line 40: Should we even wait if the retries number is reached at this point, think MAX_RETRIES=0

blockchains/eth/eth_worker.js
```diff
@@ -29,6 +30,32 @@ class ETHWorker extends BaseWorker {
...
+          continue;
+        }
+        return response;
+      } catch(err) {
+        retries++;
+        retryIntervalMs += (2000 * retries);
+        logger.error(
+          `Try block in ${params[0]} failed. Reason: ${err.toString()}. Waiting ${retryIntervalMs} and retrying for ${retries} time`
+          );
+        await new Promise((resolve) => setTimeout(resolve, retryIntervalMs));
```
YordanPavlov, line 53: This logic seems to be the same as above, can it be moved to `waitOnError()` method?

blockchains/eth/lib/next_interval_calculator.js
```diff
@@ -1,56 +1,41 @@
...
-      worker.sleepTimeMsec = 0;
-    }
-    else {
-      // If data became available only after asking the Node, we are close to the Head, come back later
-      worker.sleepTimeMsec = constants.LOOP_INTERVAL_CURRENT_MODE_SEC * 1000;
-    }
+  worker.sleepTimeMsec = 0;
+  const progressDifference = worker.lastConfirmedBlock - worker.lastExportedBlock;
+  const maxInterval = constants.MAX_CONCURRENT_REQUESTS * constants.BLOCK_INTERVAL;
+  let intervalArrayLength;
```
YordanPavlov, line 26: Can be made const with the ternary operator.

blockchains/eth/lib/next_interval_calculator.js
```diff
@@ -1,56 +1,41 @@
...
-  if (firstNewBlockCheck || isNewBlockAvailable(worker)) {
-    // If data was available without asking with Node, we are catching up and should come back straight away
-    if (firstNewBlockCheck) {
-      worker.sleepTimeMsec = 0;
-    }
-    else {
-      // If data became available only after asking the Node, we are close to the Head, come back later
-      worker.sleepTimeMsec = constants.LOOP_INTERVAL_CURRENT_MODE_SEC * 1000;
-    }
+  worker.sleepTimeMsec = 0;
```
YordanPavlov, line 23: This sleep logic seems to be much simpler here and the logic is more or less the same. Good job.

blockchains/eth/eth_worker.js
spiderjako: param[0] is the method name. 'eth_getLogs' for example

blockchains/eth/lib/next_interval_calculator.js
spiderjako: Could be, but IMO it looks cleaner this way. With ternary it would be a bit too bloated. I like to flex with ternary but I realised a few weeks back that I don't have to always do it

blockchains/eth/eth_worker.js
spiderjako: I think that it would work with an array if it works with a batch of requests. Technically, [<request1>] is a batch, but I'd have to rework how the parameters are given. Currently we have 
```js
('trace_filter', [{
  fromBlock: this.web3Wrapper.parseNumberToHex(fromBlock),
  toBlock: this.web3Wrapper.parseNumberToHex(toBlock)
}])
```

blockchains/eth/eth_worker.js
spiderjako: What do you mean?

blockchains/eth/eth_worker.js
spiderjako: I'd leave it as it is and rework both this and the UTXO version in another PR. They currently have basically the same idea. Maybe I'll even combine them. Currently, the loggings point to different errors in general, one is to the response's error, the other is a caught error

blockchains/eth/eth_worker.js
YordanPavlov: I would expect MAX_RETRIES to specify the number of retries, that is attempts to re-send after the initial sending. Instead if it is is 0, we would not send the request at all. If it is 1, we would send one time, on error wait, but then not retry. Seems to me like the logic is a bit off.

blockchains/eth/lib/next_interval_calculator.js
YordanPavlov: It's not about flexing, const variables make it easier to reason about the code. Minor ofc, your call.

YordanPavlov (approved):

YordanPavlov:
Legacy by now, if we need the feature would require full rewrite.

PR #166, opened 2023-11-28
Fix for 'Maximum call stack size exceeded'

YordanPavlov:
Using the spread operator on a huge array would throw error. Go back to good old iteration.


spiderjako (approved):
Hm, TIL
```
Its syntax is simple and easy to understand. While the spread operator is a convenient way to achieve this, it can be less performant than using a traditional loop. This is because the spread operator creates a new array or object every time it is used, which can be expensive when working with large arrays/objects
```

PR #167, opened 2023-12-06
Improving on the parallelism idea

spiderjako:
### Done
* Refactored the workLoop to work in the following manner: ⭕ 
1. Add tasks to the queue, unless it's full (constant should be better defined, currently set at 100 by default)
2. Take whatever is in the "buffer" array (maybe not a good name), sort it and take each value that has a block number in the current or the next interval. ⚠️ In the code under a TODO I've outlined a potential optimisation in regards to blocks without data that we need / empty blocks. Will be done by the end of the day (22.12)
3. Store this non-descending (by block number) array in Kafka
4. Save progress ⚠️ this would be solved with the TODO that I've outlined

* Took optimisations from this [PR](https://github.com/santiment/san-chain-exporter/pull/162) (nextIntervalCalculator mainly) ⭕ 
* Made optimisations to the order in which we fetch traces, blocks and receipts ⭕ 
* Removed async/await from some methods and renamed them for better clarity ⭕ 

### TODOs
* PR is rebased and refactored using the master code (a lot of changes were made during the last 1-2 weeks) ❌ 
* Final changes on how the progress is being stored and how to handle empty blocks (idea is in the TODOs in the code) ❌ 
* Other blockchains are refactored in order to work with the new ❌ 
* Add Tests, check existing tests, due name-changes ❌


index.js
```diff
@@ -43,6 +50,10 @@ class Main {
     this.worker = new worker.worker();
     await this.worker.init(this.exporter, metrics);
     await this.handleInitPosition();
+    if (BLOCKCHAIN === 'eth') {
```
spiderjako, line 53: ⚠️  Temporary: Currently testing with the Ethereum codebase. When done, would rework the rest of the code to work in the new way

blockchains/eth/eth_worker.js
```diff
@@ -124,54 +130,44 @@ class ETHWorker extends BaseWorker {
...
-      }
+    for (const interval of intervals) this.queue.add(() => this.makeQueueTask(interval));
 
-      this.lastPrimaryKey += events.length;
-    }
-
-    this.lastExportedBlock = result.toBlock;
-
-    return events;
+    this.lastExportedBlock = Math.max(intervals[intervals.length - 1].toBlock, this.lastExportTime);
```
spiderjako, line 170: This doesn't make sense ATM, should be removed/refactored

PR #168, opened 2023-12-07
Migrate Web3JS to modern version

YordanPavlov:
Up until now we have been using the legacy Web3JS version 1, in this PR we upgrade to version 4. As the libraries are generations apart there are some source incompatible changes.

* The main one is the new version returning `BigInt` for some numeric values. We mostly cast those back to Number as we need to serialize it to JSON.
* Some Web3 methods were more forgiving in version 1. For example `hexToNumber` would accept a non hex argument and still convert it. In the new version we need to make sure that we provide a string in the format '0x...'.

As part of the change I also did some refactoring:
* Wrap all the calls to `Web3` into `Web3Wrapper`. Up until now we did some of the calls through the `Web3Wrapper` and some directly to the `Web3` module. For consistency I think it is easier to wrap all calls.
* Dependency injection for workers. Workers now receives the constants they need as a constructor argument instead of loading the settings with `require`. This helps testing as we do not need to do some complex runtime modifications of the required constants and just provide what we need in the test.
* Minor bugs, left comments in the review.


blockchains/bnb/bnb_worker.js
```diff
@@ -36,11 +35,12 @@ class MetricsStore {
 
 
 class BNBWorker extends BaseWorker {
-  constructor() {
-    super();
+  constructor(constants) {
```
YordanPavlov, line 38: Provide the needed settings as constructor argument instead of the worker including them. This helps testing and makes more obvious what the worker needs.

blockchains/erc20/lib/fetch_events.js
```diff
@@ -13,27 +13,32 @@ const QNT_contract = '0x4a220e6096b25eadb88358cb44068a3248254675';
 const WETH_contract = '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2';
 
 
-async function decodeEventBasicInfo(web3, event, timestampsCache) {
-  const timestamp = await timestampsCache.getBlockTimestamp(web3, event['blockNumber']);
+async function decodeEventBasicInfo(web3Wrapper, event, timestampsCache, addContract = true) {
```
YordanPavlov, line 16: Add an extra argument to this function so that it can be reused by Matic codebase.

blockchains/matic/lib/fetch_events.js
```diff
@@ -2,35 +2,25 @@
 const { decodeEvents } = require('../../erc20/lib/fetch_events');
 const { TimestampsCache } = require('../../erc20/lib/timestamps_cache');
 const { decodeAddress } = require('../../erc20/lib/util');
+const { decodeEventBasicInfo } = require('../../erc20/lib/fetch_events');
```
YordanPavlov, line 5: Add one extra argument to the generic method and reuse it.

blockchains/receipts/lib/helper.js
```diff
@@ -18,49 +17,57 @@ const parseTransactionReceipts = (responses) => {
...
   const result = {};
   collection.forEach(logKeys, key => result[`logs.${key}`] = decodedLogs.map(log => log[key]));
 
   return result;
 };
 
-const decodeReceipt = (receipt) => {
-  const clonedReceipt = lang.clone(receipt);
+const decodeReceipt = (receipt, web3Wrapper) => {
+  const clonedReceipt = lang.cloneDeep(receipt);
```
YordanPavlov, line 47: Bugfix. `lang.clone` does a shallow copy and we need a deep one.

blockchains/receipts/lib/helper.js
```diff
@@ -18,49 +17,57 @@ const parseTransactionReceipts = (responses) => {
...
 
   collection.forEach(['blockNumber', 'status', 'transactionIndex'],
-    key => clonedReceipt[key] = Web3.utils.hexToNumber(clonedReceipt[key])
+    key => {
+      if (Object.prototype.hasOwnProperty.call(clonedReceipt, key) && clonedReceipt[key] !== undefined) {
```
YordanPavlov, line 51: We need to check if those fields are actually present, otherwise we would set `null`.

blockchains/bnb/bnb_worker.js
WonderBeat: it might be more convenient if we call this `settings` instead of `constants`. `this.constants` sounds a bit weird.

blockchains/cardano/cardano_worker.js
```diff
@@ -3,16 +3,16 @@
...
 
 const CARDANO_GRAPHQL_URL = process.env.CARDANO_GRAPHQL_URL || 'http://localhost:3100/graphql';
 const DEFAULT_TIMEOUT_MSEC = parseInt(process.env.DEFAULT_TIMEOUT || '30000');
 
 class CardanoWorker extends BaseWorker {
-  constructor() {
-    super();
+  constructor(constants) {
+    super(constants);
+    this.constants = constants;
```
WonderBeat, line 15: I wonder why we need this if base class already have a field with the same name. Most probably we could access it

WonderBeat (approved):

spiderjako (approved):

PR #172, opened 2023-12-13
Retry cardano requests

YordanPavlov:
In this PR we implement retry on Cardano requests, this should reduce the pod restarts.

### DIFF
--- a/blockchains/cardano/cardano_worker.js
+++ b/blockchains/cardano/cardano_worker.js
@@ -12,7 +12,6 @@ const DEFAULT_TIMEOUT_MSEC = parseInt(process.env.DEFAULT_TIMEOUT || '30000');
 class CardanoWorker extends BaseWorker {
   constructor(settings) {
     super(settings);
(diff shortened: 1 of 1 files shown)


blockchains/cardano/cardano_worker.js
```diff
@@ -12,6 +12,7 @@ const DEFAULT_TIMEOUT_MSEC = parseInt(process.env.DEFAULT_TIMEOUT || '30000');
 class CardanoWorker extends BaseWorker {
   constructor(settings) {
     super(settings);
+    this.pRetry = null;
```
spiderjako, line 15: Why not just leave it at undefined?
```js
this.pRetry;
```

spiderjako (approved):

blockchains/cardano/cardano_worker.js
YordanPavlov: I thought this gives some clarity that such member variable would be later used, but obviously it is questionable. Removed.

PR #173, opened 2023-12-15
Optimize block timestamp discovery. Integration tests.

YordanPavlov:
In this PR:
* Optimize the request to find out block timestamps so that a single batch RPC request is used.
* Add integration tests.


README.md
```diff
@@ -42,10 +42,39 @@ $ npm test
...
+```
+./e2e/start-environment.sh
+```
+
+2. Run the test itself
+
+In another console execute the test:
+
+```
+./e2e/test.sh
```
YordanPavlov, line 61: This test was not working. Resurrect it, but make it run outside of the docker compose. This helps to debug the test with a debugger or to analyze if the infrastructure is not properly initialized.

README.md
```diff
@@ -42,10 +42,39 @@ $ npm test
...
+```
+./e2e/stop-environment.sh
+```
+
+## Integration tests
+
+Those are tests which would run the exporter against a running Node and compare the output to expected values.
+
+```
+npm run integration_test
```
YordanPavlov, line 77: New type of tests. See below.

blockchains/erc20/lib/timestamps_cache.js
```diff
@@ -1,45 +1,54 @@
...
 
 class TimestampsCache {
-  constructor() {
+  constructor(ethClient, fromBlock, toBlock) {
     this.timestampStore = {};
-  }
+    this.rangeSize = toBlock - fromBlock + 1;
+
+    const blockRequests = [];
+    for (let i = fromBlock; i <= toBlock; i++) {
```
YordanPavlov, line 9: One of the two main features of this PR. In this TimestampsCache constructor we would send the batch RPC request to get all the blocks for the range we are currently fetching. What we did before was do a separate request once we need a new block, this would result in N requests if we are fetching a range of N blocks.

blockchains/erc20/lib/fetch_events.js
```diff
@@ -151,13 +151,13 @@ async function decodeWETHDepositEvent(web3Wrapper, event, timestampsCache) {
...
  **/
-async function decodeWETHWithdrawalEvent(web3Wrapper, event, timestampsCache) {
+function decodeWETHWithdrawalEvent(web3Wrapper, event, timestampsCache) {
   if (event['address'].toLowerCase() !== WETH_contract
     || event['topics'].length !== 2) {
     return null;
   }
 
-  let result = await decodeEventBasicInfo(web3Wrapper, event, timestampsCache);
+  const result = decodeEventBasicInfo(event, timestampsCache);
```
YordanPavlov, line 160: The `timestampsCache` would already contain the data we need (see below). We do not do an async call to fill it anymore at this point. As result we do not 'await'.

e2e/docker-compose.yml
```diff
@@ -20,33 +20,23 @@ services:
...
     logging:
       driver: none
-  exporter:
-    build:
-      context: ../
-      dockerfile: docker/Dockerfile
-      args:
-        NODE_ENV: development
+  kafka-init:
+    image: wurstmeister/kafka
```
YordanPavlov, line 38: For this test we need one kakfa topic to exist. We used to do it inside the test but it complicates the logic there. Now do it with an extra container here.

e2e/producer-transaction.spec.js
```diff
@@ -77,45 +67,13 @@ class TestConsumer {
...
-  await exporter.commitTransaction();
-}
 
-describe('Producer transactions', function() {
+describe('Producer transactions', function () {
   let exporter;
   let testConsumer;
   let num_messages_test = 3;
 
-  // This is done only to create the test topic.
```
YordanPavlov, line 100: Not anymore.

e2e/test.sh
```diff
@@ -1,3 +1,3 @@
- #! /bin/sh
+#! /bin/sh
 
-docker-compose -f ./e2e/docker-compose.yml up --build && docker-compose -f ./e2e/docker-compose.yml rm -f
+KAFKA_URL=localhost:9092 ZOOKEEPER_URL=localhost:2181 KAFKA_TOPIC=erc20_exporter_test ./node_modules/.bin/mocha e2e/producer-transaction.spec.js
```
YordanPavlov, line 3: Env variables used to come from the docker-compose, now we set them here.

integration_test/src/binance-native-token.spec.js
```diff
@@ -0,0 +1,36 @@
+const assert = require('assert');
+const worker = require('../../blockchains/eth/eth_worker');
+
+
+describe('BEP20 worker test', function () {
+  it('BEP20 worker should extract blocks from 20000000 to 20000999 including', async function () {
```
YordanPavlov, line 6: Second feature in this PR. Add the skeleton for what should become a set of integration tests for each chain. We query a range of blocks from the running Node and compare with how we expect this data to look like.

README.md
WonderBeat: I can see no way to maintain this test if we don't run it in Jenkins. Can we run it every build?

blockchains/erc20/lib/fetch_events.js
```diff
@@ -13,8 +13,8 @@ const QNT_contract = '0x4a220e6096b25eadb88358cb44068a3248254675';
 const WETH_contract = '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2';
 
 
-async function decodeEventBasicInfo(web3Wrapper, event, timestampsCache, addContract = true) {
```
WonderBeat, line 16: 👍🏻 for removing `async`! Well done!

blockchains/erc20/lib/timestamps_cache.js
WonderBeat: image

blockchains/erc20/lib/timestamps_cache.js
```diff
@@ -1,45 +1,54 @@
...
-    this.timestampStore[blockNumber] = timestamp;
+    this.responsePromise = ethClient.request(blockRequests);
   }
 
-  getTimestampFromStore(blockNumber) {
-    if (Object.prototype.hasOwnProperty.call(this.timestampStore, blockNumber)) {
-      return this.timestampStore[blockNumber];
+  async waitResponse(web3Wrapper) {
+    const resultsArray = await this.responsePromise;
+    if (!Array.isArray(resultsArray)) {
```
WonderBeat, line 25: And then they come and hate Java verbosity ;)

blockchains/erc20/lib/timestamps_cache.js
```diff
@@ -1,45 +1,54 @@
...
+      throw new Error('Blocks response is not an array');
+    }
+    if (resultsArray.length !== this.rangeSize) {
+      throw new Error(`Expected ${this.rangeSize} but got ${resultsArray.length} blocks response`);
     }
 
-    return DATA_MISSING;
-  }
+    for (const result of resultsArray) {
```
WonderBeat, line 32: There are libraries like underscore.js (old one) which can bring style here. `map, filter, fold` guys are waiting to improve readability

WonderBeat (approved):

blockchains/erc20/lib/timestamps_cache.js
YordanPavlov: I thought it's a classic we all like, no? :) You would suggest descriptive names like 'counter'?

README.md
spiderjako: Why would you run integration tests with every build? Isn't that reserved for unit tests?
--
On the PR side, I saw the `e2e` folder a while ago and realised quite later that those were some tests. Didn't give it much thought as it didn't work, but now that we're on it - shouldn't all tests be in the `test` folder and that folder should be separated into `unit` and `integration`? Gives more clarity to the structure IMO.

spiderjako (approved):
Minor comment - binance native token is BSC, your test should be just BEP20 or something similar

README.md
YordanPavlov: I am not sure we need that test to run all the time, it is more of a test of node-rdkafka and demo how to use transactions. I added this code back when I added a [PR](https://github.com/Blizzard/node-rdkafka/pull/881) for transactions support in node-rdkafka.

antisaa (approved):

blockchains/erc20/lib/timestamps_cache.js
YordanPavlov: Nowadays functional appraoch is available through the language itself, ex:

```
const newTimestampStore = resultsArray.reduce((timestampStore, result) => {
  const blockNumber = web3Wrapper.parseHexToNumber(result.result.number);
  const blockTimestamp = web3Wrapper.parseHexToNumber(result.result.timestamp);
  return {
    ...timestampStore,
    [blockNumber]: blockTimestamp
  };
}, {});
```
However even if we presume that this is better as there are no 'side effects' the performance is worse. The spred operator would construct an extra object on each iteration. Also spread operator would throw SO exception on big objects.

PR #175, opened 2023-12-29
Increase TCP session inactivity timeout from 5 to 30 seconds

YordanPavlov:
It turned out that the TCP session constructed by Node HTTP module would have an inactivity timeout of 5 seconds by default. To prove this observation I constructed a simple client-server application:

The server:

```js
const http = require('http');

const server = http.createServer((req, res) => {
  if (req.method === 'POST') {
    let body = '';

    req.on('data', chunk => {
      body += chunk.toString(); // Convert binary chunk to string
    });

    req.on('end', () => {
      try {
        const jsonRpcRequest = JSON.parse(body);

        if (jsonRpcRequest.method === 'multiply' && Array.isArray(jsonRpcRequest.params)) {
          // Perform the multiplication
          const result = jsonRpcRequest.params[0] * jsonRpcRequest.params[1];
          const response = {
            jsonrpc: '2.0',
            id: jsonRpcRequest.id,
            result: result
          };
          res.writeHead(200, { 'Content-Type': 'application/json' });
          res.end(JSON.stringify(response));
        } else {
          // Invalid method or params
          res.writeHead(400);
          res.end('Invalid JSON-RPC method or params');
        }
      } catch (error) {
        // Error parsing JSON or handling the request
        res.writeHead(500);
        res.end('Server error: ' + error.message);
      }
    });
  } else {
    // Not a POST request
    res.writeHead(405); // Method Not Allowed
    res.end();
  }
});

server.listen(3000, () => {
  console.log('Server listening on http://localhost:3000');
});
```

The client:

```js
const http = require('http');

// A function to send JSON-RPC requests
function sendJsonRpcRequest(method, params, id, callback) {
  const data = JSON.stringify({
    jsonrpc: '2.0',
    method: method,
    params: params,
    id: id
  });

  const options = {
    hostname: 'localhost',
    port: 3000,
    path: '/',
    method: 'POST',
    headers: {
      'Content-Type': 'application/json',
      'Content-Length': Buffer.byteLength(data)
    }
  };

  const req = http.request(options, res => {
    res.setEncoding('utf8');
    let responseBody = '';

    res.on('data', chunk => {
      responseBody += chunk;
    });

    res.on('end', () => {
      try {
        const response = JSON.parse(responseBody);
        callback(null, response);
      } catch (error) {
        callback(error);
      }
    });
  });

  req.on('error', callback);
  req.write(data);
  req.end();
}

// A client request to the multiply method
sendJsonRpcRequest('multiply', [5, 5], 1, (err, response) => {
  if (err) {
    console.error('Error:', err);
    return;
  }

  console.log('Result:', response.result); // 25
});

setTimeout(() => {
  console.log('Time is up. Exit.');
}, 60000);

```

The network communication of this application is this:

image

As you can see there are 5 keep alive packets sent with 1 second interval between them, and then the connection is closed. This behavior affects the Jayson library we use, which is a thin wrapper of the Node HTTP module, providing some RPC functionality.

### DIFF
--- a/lib/http_client.js
+++ b/lib/http_client.js
@@ -1,0 +1,38 @@
+const http = require('http');
+const https = require('https');
+const jayson = require('jayson/promise');
+
+// The TCP session constructed by Node HTTP module would get closed after 5 seconds of inactivity by default.
+// Extend this timeout to 30 to reduce the number of sessions constructed.
+const TCP_SESSION_KEEP_ALIVE_MSEC = 30000;
+
+function constructRPCClient(nodeURL, extraOptions = {}) {
+    const nodeUrl = new URL(nodeURL);
+
+    const mergedOptions = {
+        hostname: nodeUrl.hostname,
+        port: nodeUrl.port,
+        path: nodeUrl.pathname,
+        ...extraOptions
+    };
+
+    const agentOptions = {
+        keepAlive: true, // Enable keep-alive
+        keepAliveMsecs: TCP_SESSION_KEEP_ALIVE_MSEC // Keep alive for 30 seconds
+    };
+
+    if (nodeURL.substring(0, 5) === 'https') {
+        const agent = new https.Agent(agentOptions);
+        mergedOptions.agent = agent;
+        return jayson.client.https(mergedOptions);
+    } else {
+        const agent = new http.Agent(agentOptions);
+        mergedOptions.agent = agent;
+        return jayson.client.http(mergedOptions);
+    }
+}
+
+
+module.exports = {
+    constructRPCClient: constructRPCClient
+};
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -33,7 +33,6 @@
     "prom-client": "^13.1.0",
     "segfault-handler": "^1.3.0",
     "web3": "^4.2.2",
-    "whatwg-url": "^11.0.0",
```
YordanPavlov, line 36: Not needed in modern Node versions

lib/http_client.js
```diff
@@ -0,0 +1,38 @@
...
+    } else {
+        const agent = new http.Agent(agentOptions);
+        mergedOptions.agent = agent;
+        return jayson.client.http(mergedOptions);
+    }
+}
+
+
+module.exports = {
+    constructRPCClient: constructRPCClient
```
spiderjako, line 37: No need for doubling, could just be `constructRPCClient`

spiderjako (approved):

WonderBeat (approved):

PR #178, opened 2024-01-07
Abstract the parallelism idea into a single entity

spiderjako:
### TaskManager
* After a discussion I've had with @WonderBeat I decided that it's a good idea to abstract the whole [parallelism](https://github.com/santiment/san-chain-exporter/pull/167) functionality into a single entity, as much as I can, with minimal touches to the main and worker code. Also, I tried to break the aforementioned PR into several ones with a few commits instead of the monster that it was. 
* Also @YordanPavlov gave me a good idea on how the data should be pushed into the buffer, using a mapping for the intervals
* TaskManager, because of the whole task queue thing, had some nostalgia for Windows for some reason. The name could be changed
* It basically takes care of 2 things - pushing tasks to a queue and pushing the done tasks from the mapping to the buffer, which will be used for storing data into Kafka


index.js
```diff
@@ -75,19 +84,27 @@ class Main {
...
-      const events = await this.worker.work();
-
       this.worker.lastExportTime = Date.now();
 
       this.updateMetrics();
       this.lastProcessedPosition = this.worker.getLastProcessedPosition();
 
-      if (events && events.length > 0) {
-        await this.exporter.storeEvents(events);
+      if (this.taskManager.buffer && this.taskManager.buffer.length > 0) {
```
YordanPavlov, line 101: Currently the consumer of the buffer is part of the main loop. I think it would make sense to separate it. Make the buffer blocking (wrap it into a simple class where you can call `await waitOnData()`) and then start the consumer and make it block on the buffer.

index.js
```diff
@@ -75,19 +84,27 @@ class Main {
 
   async workLoop() {
     while (this.shouldWork) {
+      if (this.taskManager.queue.size < constantsBase.PQUEUE_MAX_SIZE) {
+        const interval = await nextIntervalCalculator(this.worker);//TODO:
+        if (interval.success) {
+          delete interval.success;
+          this.taskManager.pushToQueue(() => this.worker.work(interval));
+          this.worker.lastExportedBlock = interval.toBlock;
```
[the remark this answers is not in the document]
spiderjako: lastExportedBlock is only used for nextIntervalCalculator, meaning, it just gets its intervals from this
> the work is not done yet, just scheduled
indeed, the tasks are just pushed with the calculated intervals, there's no guarantee, that the last one is finished, but lastExportedBlock doesn't overwrite any ZK state

index.js
```diff
@@ -29,11 +31,17 @@ class Main {
...
     this.lastProcessedPosition = this.worker.initPosition(lastRecoveredPosition);
     await this.exporter.savePosition(this.lastProcessedPosition);
   }
+  
+  async #initTaskManager(blockNumber) {
+    this.taskManager = new TaskManager(blockNumber, constants.BLOCK_INTERVAL);
+    await this.taskManager.initPQueue(MAX_CONCURRENT_REQUESTS);
+  }
 
   #isWorkerSet() {
```
WonderBeat, line 46: it's quite common for `isSomething` methods to return `bool` as an answer. This one returns void and confuses me a bit

index.js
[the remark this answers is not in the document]
WonderBeat: Then I suggest rename this field. `lastExportedBlock` is not what it seems.

index.js
```diff
@@ -75,19 +84,27 @@ class Main {
...
 
       this.updateMetrics();
       this.lastProcessedPosition = this.worker.getLastProcessedPosition();
 
-      if (events && events.length > 0) {
-        await this.exporter.storeEvents(events);
+      if (this.taskManager.buffer && this.taskManager.buffer.length > 0) {
+        await this.exporter.storeEvents(this.taskManager.buffer);
+        this.lastProcessedPosition = this.taskManager.lastProcessedPosition;
+        this.taskManager.buffer = [];
```
WonderBeat, line 104: I'm worried and almost sure that buffer may be modified by Task Manager between `storeEvents` and `clean`

lib/task_manager.js
```diff
@@ -0,0 +1,49 @@
...
+    this.queue;
+    this.buffer = [];
+    this.blockInterval = blockInterval;
+    this.currentInterval = {
+      fromBlock: startBlock + 1,
+      toBlock: startBlock + blockInterval
+    };
+  }
+
+  async initPQueue(maxConcurrentRequests) {
```
WonderBeat, line 13: We should not expose internal implementation to client. PQueue is too specific. Could you change method name?

WonderBeat (changes requested):

index.js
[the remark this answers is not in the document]
YordanPavlov: > lastExportedBlock is only used for nextIntervalCalculator
It is meant to be exposed to Prometheus, see `updateMetrics()`

index.js
```diff
@@ -75,19 +84,27 @@ class Main {
 
   async workLoop() {
     while (this.shouldWork) {
+      if (this.taskManager.queue.size < constantsBase.PQUEUE_MAX_SIZE) {
```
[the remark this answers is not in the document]
spiderjako: I think current implementation is better, the idea is to continue with kafka pushing, even if the queue is full. If I await the promise, I block the program, at least, as far as I understand
`Returns a promise that settles when the queue size is less than the given limit: queue.size < limit.`

lib/task_manager.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+  }
+
+  handleNewData(newTransformedData, interval) {
+    this.map[JSON.stringify(interval)] = newTransformedData;
+    this.#pushAllEligable();
+  }
+
+  pushToQueue(task) {
+    this.queue.add(task);
```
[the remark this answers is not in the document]
spiderjako: I'd disagree. The current thing is as simple and light as it gets, and making the queue elsewhere would just create a more complex web of dependent code. I'd say this is good as a single entity - queue of tasks, which push data to a buffer when done.

index.js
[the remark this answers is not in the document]
YordanPavlov: Just separate the things, and do not do them in a single loop. If you go back from this loop and wait for some magic number of milliseconds you are reducing the performance of your program. Also the Kafka writing, again do not do it in this loop.

lib/task_manager.js
```diff
@@ -0,0 +1,67 @@
+class TaskManager {
+  constructor() {
+    this.map = {};
+    this.queue;
+    this.buffer = [];
+    this.lastPushedToQueue = 0;
+  }
+
+  async initQueue(maxConcurrentRequests) {
```
WonderBeat, line 9: This might be called in the scope of constructor. as a result we'll get ready to use object

lib/task_manager.js
```diff
@@ -0,0 +1,67 @@
...
+    const PQueue = (await import('p-queue')).default;
+    this.queue = new PQueue({ concurrency: maxConcurrentRequests });
+    this.queue.on('completed', (data) => this.handleNewData(data));
+    // this.queue.on('error', err => { return Promise.reject(err.toString()); } );
+  }
+
+  clearQueue() {
+    this.taskManager.queue.pause();
```
WonderBeat, line 17: pause doesn't pause already in flight tasks. We clean queue next step which won't work if there's a long running task running

lib/task_manager.js
```diff
@@ -0,0 +1,67 @@
...
+    this.taskManager.queue.clear();
+  }
+
+  #generateDefensiveCopy() {
+    const bufferCopy = [];
+    for (const data of this.buffer) bufferCopy.push(data);
```
WonderBeat, line 23: most probably we can use JS clone which should be faster and more readable

lib/task_manager.js
```diff
@@ -0,0 +1,67 @@
+class TaskManager {
+  constructor() {
+    this.map = {};
```
WonderBeat, line 3: there should be a better name for this var

lib/task_manager.js
```diff
@@ -0,0 +1,67 @@
...
+        };
+      }
+
+      delete this.map[this.lastPushedToQueue];
+      this.lastPushedToQueue++;
+    }
+  }
+
+  handleNewData(newTransformedData) {
+    this.map[this.lastPushedToQueue] = newTransformedData;
```
WonderBeat, line 58: hmmm. Maybe I'm wrong but I think this doesn't work if we have our tasks completed out of order. i.e. [3,2,1] - this will update lastProcessedPosition to blockNumber `1`. right?

lib/task_manager.js
spiderjako: p-queue can't be initialised in the constructor

blockchains/eth/eth_worker.js
```diff
@@ -40,7 +39,9 @@ class ETHWorker extends BaseWorker {
     return this.ethClient.request('trace_filter', [{
       fromBlock: this.web3Wrapper.parseNumberToHex(fromBlock),
       toBlock: this.web3Wrapper.parseNumberToHex(toBlock)
-    }]).then((data) => this.parseEthInternalTrx(data['result']));
+    }])
+      .then((data) => this.parseEthInternalTrx(data['result']))
+      .catch((err) => { throw err; });
```
YordanPavlov, line 44: Redundant

blockchains/eth/eth_worker.js
```diff
@@ -145,11 +146,10 @@ class ETHWorker extends BaseWorker {
     return result;
   }
 
-  async work() {
+  async work(key) {
```
YordanPavlov, line 149: You give this 'key' argument to the `work()` method just to have it back. It does not need to know about this key, it does nothing with it. Wrap the call in a closure if you need to preserve it.

blockchains/eth/lib/next_interval_calculator.js
```diff
@@ -1,53 +1,29 @@
...
-    // The Node has not progressed
     worker.sleepTimeMsec = worker.settings.LOOP_INTERVAL_CURRENT_MODE_SEC * 1000;
     return { success: false };
   }
+
+  worker.sleepTimeMsec = 0;
+  return {
+    success: true,
+    fromBlock: worker.lastBufferedBlock + 1,
+    toBlock: Math.min(worker.lastBufferedBlock + worker.settings.BLOCK_INTERVAL, worker.lastConfirmedBlock)
```
YordanPavlov, line 25: This function depends on some side effects to work correctly. The worker must modify `worker.lastBufferedBlock` before the next call, this is imperative or same intervals would be created. I think the logic of the function can be improved in two ways:
1. Break the 'isNewBlockAvailable' logic. It has nothing to do with the interval.
2. Make the arguments which the function depend on more explicit. This would make it more obvious to the caller that it is up to him to provide updated 'from' value. The function needs to know two arguments (lastConfirmedBlock, toBlock).

lib/task_manager.js
```diff
@@ -0,0 +1,66 @@
...
+
+  #generateDefensiveCopy() {
+    const bufferCopy = [];
+    for (const data of this.buffer) bufferCopy.push(data);
+    return bufferCopy;
+  }
+
+  #shiftOriginalBuffer(shiftsCount) {
+    while (shiftsCount >= 0) {
+      this.buffer.shift();
```
YordanPavlov, line 28: What's that shenanigan with shifting by one. Just erase the whole buffer by setting length to 0.

lib/worker_base.js
```diff
@@ -43,7 +43,7 @@ class WorkerBase {
    */
   getLastProcessedPosition() {
     return {
-      blockNumber: this.lastExportedBlock,
+      blockNumber: this.lastBufferedBlock,
```
YordanPavlov, line 46: That's wrong. This is the state we would preserve in ZK from which we would recover on restart. Last buffered block is too optimistic, this is not yet in Kafka.

lib/task_manager.js
```diff
@@ -0,0 +1,66 @@
...
+    this.#shiftOriginalBuffer(shiftsCount);
+
+    return bufferCopy;
+  }
+
+  #pushAllEligable() {
+    while (this.taskData[this.lastPushedToBuffer]) {
+      for (const data of this.taskData[this.lastPushedToBuffer]) {
+        this.buffer.push(data);
+        this.lastProcessedPosition = {
```
YordanPavlov, line 45: This is getting messy, we have 'lastProcessed' and 'lastBuffered'. However here you are mixing them here, the fact that something is eligible for going in the buffer doesn't make it any more 'processed', it still needs to be consumed from the Kafka side and send through the producer.

YordanPavlov: You don't need this member variable. The user of this class can see what the last value is once it calls `retrieveCompleted`

blockchains/cardano/cardano_worker.js
```diff
@@ -170,7 +170,7 @@ class CardanoWorker extends BaseWorker {
   }
 
   async work() {
-    const fromBlock = this.lastExportedBlock + 1;
+    const fromBlock = this.lastBufferedBlock + 1;
```
YordanPavlov, line 173: Not a fan of all those renames. You resorted to renaming the 'last exported block' to 'last buffered block' to make it comply more easily with your change. But:
1. All those other workers don't know what 'buffered block' is
2. I would argue that 'last exported block' is cleaner in terms of semantics. The last block which got into Kafka. What is 'last buffered block'?

index.js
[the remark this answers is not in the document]
YordanPavlov: At the end of this loop there is a timeout (L 112) what should the timeout be when the above condition is false? You either set it to 0 and in effect busy loop until Pqueue resolves the promises or invent some number which would reduce your performance.

blockchains/eth/eth_worker.js
spiderjako: It doesn't look really elegant atm, but I didn't think of another way to somehow set an order of the tasks beforehand
---
but I think I've thought of a way with queue.on('add'), which would be elegant

lib/task_manager.js
spiderjako: If you do that, you risk losing data. 
Let's say we have a buffer of [1,2,3,4]. You copy this. In the meantime, between the copy and the buffer deletion, another element is added at the back, let's say 5. Then you'll erase the whole buffer
Though, I think this could be improved with Array.slice, now that you say it. Maybe it'd be a bit more optimal. But a shenanigan is indeed needed

blockchains/eth/eth_worker.js
YordanPavlov: > Wrap the call in a closure if you need to preserve it.

Wouldn't what I suggested work? I am thinking this:

```js
const key = "some key"
this.taskManager.pushToQueue(() => {
  const result = this.worker.work()
  return [key, result]
  }
);
  ```

File: lib/task_manager.js (hunk 16 above)
Comment (YordanPavlov) [reply lib/task_manager.js:28]: >  In the meantime, between the copy and the buffer deletion, another element is added at the back

What do you mean, NodeJS is single threaded, this is the whole proposition. Also if it wasn't (we were dealing with some other language) this approach to access the same data structure most probably wouldn't have been enough to give us thread safety.

File: blockchains/eth/eth_worker.js (hunk 14 above)
Comment (spiderjako) [reply blockchains/eth/eth_worker.js:149]: Hm, this is simple, yes. Thanks

File: lib/task_manager.js [hunk 20]
```diff
@@ -0,0 +1,50 @@
...
+
+  handleNewData([key, newTransformedData]) {
+    this.taskData[key] = newTransformedData;
+    this.#pushAllEligable();
+  }
+
+  pushToQueue(workTask) {
+    this.queue.add(() => {
+      const result = workTask();
+      const currIndex = this.taskIndex;
```
spiderjako, line 43: This I thought would be enough, but will update this. Seems this is not enough for copying.

lib/task_manager.js
WonderBeat: Why? what's wrong with this?

lib/task_manager.js
```diff
@@ -0,0 +1,80 @@
...
+   * On the completion of a task in the p-queue, the task
+   * should return an array in the form of [interval, events].
+   * These would be set up in the taskData object accordingly in the
+   * correct format, after which we use the #pushAllEligable private method
+   * to push ready-to-go sequential data into the buffer.
+   * @param {object} interval
+   * @param {Array} newTransformedData 
+   */
+  handleNewData(interval, newTransformedData) {
+    this.taskData[interval.fromBlock] = { toBlock: interval.toBlock, data: newTransformedData };
```
YordanPavlov, line 66: I think at this point we need to have some blocking mechanism to block the producers. Currently we have no way to throttle the tasks coming from the PQueue and we would run into an OOM if the consumer of this data (the logic writing into Kafka) is slower. Right?

lib/constants.js
```diff
@@ -1,17 +1,23 @@
 const BLOCKCHAIN = process.env.BLOCKCHAIN;
 const CONFIG_PATH = process.env.CONFIG_PATH;
+const START_BLOCK = parseInt(process.env.START_BLOCK || '0') - 1;
 const EXPORT_BLOCKS_LIST = process.env.EXPORT_BLOCKS_LIST || false;
+const PQUEUE_MAX_SIZE = parseInt(process.env.PQUEUE_MAX_SIZE || '100');
+const MAX_CONCURRENT_REQUESTS = parseInt(process.env.MAX_CONCURRENT_REQUESTS || '1');
```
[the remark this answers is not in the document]
spiderjako: The `max size` one is for, how much tasks you could push into the queue. `max concurrent` one is for, how many tasks can work at the same time. You could have 100 of the first and 5 of the second

index.js
```diff
@@ -73,21 +80,30 @@ class Main {
...
 
-      this.updateMetrics();
       this.lastProcessedPosition = this.worker.getLastProcessedPosition();
-
-      if (events && events.length > 0) {
-        await this.exporter.storeEvents(events);
-      }
-      await this.exporter.savePosition(this.lastProcessedPosition);
-      logger.info(`Progressed to position ${JSON.stringify(this.lastProcessedPosition)}, last confirmed Node block: ${this.worker.lastConfirmedBlock}`);
+      if (this.taskManager.buffer.length > 0) await this.waitOnStoreEvents();
```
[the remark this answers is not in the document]
spiderjako: I think you can't push [] to Kafka. Initially I forgot about this check, but then got some errors.
https://github.com/santiment/san-chain-exporter/blob/master/index.js#L86
We have it in master as well. 
But maybe I'll move it into the function, as you say

lib/constants.js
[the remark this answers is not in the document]
YordanPavlov: I don't think you benefit from having two different sizes in your current design, because of how you queue new tasks - right away when the size drops. But anyways, don't see a harm in having both.

index.js
```diff
@@ -95,6 +128,26 @@ class Main {
     }
   }
 
+  async workLoopV2() {
```
YordanPavlov, line 131: I like this idea with the gradual migration.

index.js
```diff
@@ -70,7 +82,28 @@ class Main {
...
-    metrics.lastExportedBlock.set(this.worker.lastExportedBlock);
+    if (BLOCKCHAIN === 'eth') {
+      metrics.lastExportedBlock.set(this.taskManager.lastExportedBlock);
+    } else {
+      metrics.lastExportedBlock.set(this.worker.lastExportedBlock);
+    }
+  }
+
+  async waitOnStoreEvents() {
+    const buffer = this.taskManager.retrieveCompleted();
```
YordanPavlov, line 93: In your current implementation some logic which is not native for the TaskManager has been put inside it (last exported block, primary keys). It is better to have more separation of concern. For example setting primary keys as per the logic of each blockchain is not native to it, ideally the TaskManager shouldn't even know which blockchain it is working on. Decorating the result data is something which you can do here.

lib/task_manager.js
```diff
@@ -0,0 +1,182 @@
...
+    if (lastProcessedPosition) {
+      logger.info(`Resuming export from position ${JSON.stringify(lastProcessedPosition)}`);
+    } else {
+      lastProcessedPosition = {
+        blockNumber: START_BLOCK,
+        primaryKey: START_PRIMARY_KEY
+      };
+      logger.info(`Initialized exporter with initial position ${JSON.stringify(lastProcessedPosition)}`);
+    }
+    this.lastExportedBlock = lastProcessedPosition.blockNumber;
```
YordanPavlov, line 159: No need for that logic to be encapsulated into the TaskManager

lib/task_manager.js
```diff
@@ -0,0 +1,182 @@
...
+    }
+  }
+
+  /**
+   * Helper method for when we push the `taskData` events' data into the buffer
+   * @param {Array} events Events data that we get from the current key in the `taskData` property
+   * @param {Array} buffer The array that at the end of the primary function would result in the 
+   * combined data, accordingly updated with the primary keys
+   */
+  #updatePrimaryKeysPushToBuffer(events, buffer) {
```
YordanPavlov, line 89: No need for this to happen here. The TaskManager can provide the data as it is and it can be decorated on the upper level.

index.js
```diff
@@ -70,7 +95,41 @@ class Main {
...
+      metrics.lastExportedBlock.set(this.worker.lastExportedBlock);
+    }
+  }
+
+  /**
+   * Helper method for checking the blockchain in order to use the correct primary key functionality.
+   * @param {Array} buffer The array of events that would be accordingly updated with the primary keys.
+   */
+  applyPrimaryKeyLogic(buffer) {
+    if (BLOCKCHAIN === 'eth') {
```
YordanPavlov, line 110: You can leverage polymorphism and have 'decorateWithPrimaryKeys' method in different workers, instead of moving blockchain specific logic in the `index.js`.

lib/task_manager.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  /**
+   * @param {number} bufferLength The length of the buffer of updated events
+   */
+  updateLastPrimaryKey(bufferLength) {
+    this.lastPrimaryKey += bufferLength;
+  }
+
+  isChangedLastExportedBlock(lastProcessedBlock) {
```
YordanPavlov, line 108: A bit of a weird method to compare two numbers. The more natural way would be to have a simple get method here and compare at the caller. Minor, your call.

blockchains/eth/lib/next_interval_calculator.js
```diff
@@ -49,11 +49,24 @@ function nextIntervalCalculator(worker) {
   };
 }
 
+/**
+ * Function for calculating the next interval to be used in the worker's methods for querying the node.
+ * @param {BaseWorker} worker A worker instance, inherriting the BaseWorker class.
+ * @returns {object} The interval, derived from the progress of the worker
+ */
+function nextIntervalCalculatorV2(worker) {
```
YordanPavlov, line 57: Changing the original:

```js
function nextIntervalCalculator(worker) {
  return {
    fromBlock: worker.lastExportedBlock + 1,
    toBlock: Math.min(worker.lastExportedBlock + worker.settings.BLOCK_INTERVAL, worker.lastConfirmedBlock)
  };
}
```
to
```js
function nextIntervalCalculator(lastExported, lastConfirmed, blockInterval) {
...
 }
```
would allow you to reuse the same code. Also this is considered better design, as it more clearly denotes the intent of the function.

index.js
```diff
@@ -29,9 +32,22 @@ class Main {
       .catch((err) => { throw new Error(`${INIT_EXPORTER_ERR_MSG}${err.message}`); });
   }
 
+  initLastProcessedPosition(lastRecoveredPosition) {
```
YordanPavlov, line 35: This logic used to be part of the worker, what's the gain in moving it here? It seems like half the logic has remained there and half is here.

index.js
```diff
@@ -70,7 +95,41 @@ class Main {
     metrics.currentBlock.set(this.worker.lastConfirmedBlock);
     metrics.requestsCounter.inc(this.worker.getNewRequestsCount());
     metrics.requestsResponseTime.observe(new Date() - this.worker.lastRequestStartTime);
-    metrics.lastExportedBlock.set(this.worker.lastExportedBlock);
+    if (BLOCKCHAIN === 'eth') {
```
YordanPavlov, line 98: Not sure why this logic can't be unified. The 'eth' case can still have `this.worker.lastExportedBlock` set to the actual block that is written in Kafka.

lib/task_manager.js
```diff
@@ -0,0 +1,113 @@
...
+    this.lastExportedBlock = lastProcessedPosition.blockNumber;
+    this.lastPrimaryKey = lastProcessedPosition.primaryKey;
+  }
+
+  /**
+   * Gets the object, made by taking the lastExportedBlock and lastPrimaryKey variables.
+   * Used when updating the position of the exporter in ZooKeeper.
+   * @returns An object of the last exported block and last used primary key
+   */
+  getLastProcessedPosition() {
```
YordanPavlov, line 94: Why is this here, you don't need neither values in the TM?

lib/task_manager.js
```diff
@@ -0,0 +1,113 @@
...
+    return {
+      blockNumber: this.lastExportedBlock,
+      primaryKey: this.lastPrimaryKey
+    };
+  }
+
+  /**
+   * @param {number} bufferLength The length of the buffer of updated events
+   */
+  updateLastPrimaryKey(bufferLength) {
```
YordanPavlov, line 104: As mentioned before TM doesn't need to know about primary keys.

test/lib/task_manager.spec.js
```diff
@@ -0,0 +1,179 @@
...
+  it('queue is paused when `consequentTaskIndex` crosses the borderline', async () => {
+    const CONCURRENCY = 1;
+    const taskManager = await TaskManager.create(CONCURRENCY);
+    taskManager.consequentTaskIndex = MAX_TASK_DATA_KEYS;
+
+    const exampleEventsData = [ 1, 2, 3, 4, 5 ];
+    const emmitedTaskData = [{ fromBlock: 1, toBlock: 10 }, exampleEventsData];
+    taskManager.queue.emit('completed', emmitedTaskData);
+
+    expect(taskManager.queue.isPaused).to.eq(true);
```
YordanPavlov, line 59: I don't think that test reflects the implementation, I think all the TaskManager tests need revisiting. In the current implementation the TaskManager is a nice standalone implementation which can be black box tested.

test/eth/worker.spec.js
```diff
@@ -6,67 +6,49 @@ const constants = require('../../blockchains/eth/lib/constants');
...
-
-        feeResultWithPrimaryKey = v8.deserialize(v8.serialize(feeResult));
-        feeResultWithPrimaryKey.primaryKey = 1;
-
-        callResultWithPrimaryKey = v8.deserialize(v8.serialize(callResult));
-        callResultWithPrimaryKey.primaryKey = 2;
-    });
-
-
-    it('test primary key assignment', async function () {
```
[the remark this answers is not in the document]
spiderjako: It was because of my previous version, that delegated the primary key assignment elsewhere. I'll bring it back, though probably reworked, as now it's not done in the `work()`

test/index.test.js
```diff
@@ -139,39 +144,27 @@ describe('Main', () => {
...
     const mainInstance = new Main();
+    mainInstance.taskManager = await TaskManager.create(1);
     mainInstance.exporter = new Exporter('test-exporter');
     sinon.stub(worker.prototype, 'init').resolves();
-    sinon.stub(mainInstance, 'handleInitPosition').resolves();
+    mainInstance.lastProcessedPosition = { blockNumber: 10, primaryKey: 1 };
 
     await mainInstance.initWorker();
-    assert(mainInstance.handleInitPosition.calledOnce);
+    expect(mainInstance.worker).to.be.instanceof(worker);
```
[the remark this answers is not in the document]
spiderjako: handleInitPosition isn't currently called inside initWorker

index.js
```diff
@@ -73,6 +86,24 @@ class Main {
     metrics.lastExportedBlock.set(this.worker.lastExportedBlock);
   }
 
+  async waitOnStoreEvents(buffer) {
+    if (buffer.length > 0) {
+      this.worker.decorateWithPrimaryKeys(buffer);
```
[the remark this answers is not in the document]
WonderBeat: this is a bit strange that we modify function argument. could we use return value to make this code a bit easier to read?

lib/task_manager.js
```diff
@@ -0,0 +1,84 @@
+class TaskManager {
+  constructor() {
```
WonderBeat, line 2: could you make it private?

lib/task_manager.js
```diff
@@ -0,0 +1,84 @@
...
+   * Takes a `() => worker.work()` function and pushes it
+   * into the TaskManager's p-queue.
+   * @param {Function} workTask 
+   */
+  pushToQueue(taskMetadata) {
+    this.lastTaskIndex++;
+    this.taskData[this.lastTaskIndex] = taskMetadata;
+    const taskIndex = this.lastTaskIndex;
+    this.queue.add(async () => {
+      const result = await taskMetadata.lambda(taskMetadata.interval);
```
WonderBeat, line 78: can we pass `taskMetadata` to labda as is. I don't think this method should be aware of `.interval`

lib/task_manager.js
spiderjako: To make it private, I'd need to migrate this to Typescript, but I'd do that in a following PR. JS doesn't allow private classes

lib/task_manager.js
spiderjako: But you do need them. You'd need to know, which is the last handled lastExportedBlock

YordanPavlov (approved):
LGTM, let's do some integration tests before merging.

YordanPavlov:
Did not got merge at the time and would need full re-visit nowadays.

PR #179, opened 2024-01-25
Removing non existing rewards that node returns after the merge

antisaa:
On block #15537393 the rewards minting moved from L1 to Beacon chain. Erigon node returns a reward of 2 ether that is minted on L1. This change just ignores such records from the mentioned block.


spiderjako (approved):

YordanPavlov (approved):

YordanPavlov:
This change is implemented in a newer PR:
https://github.com/santiment/san-chain-exporter/pull/206

PR #180, opened 2024-01-28
Updating next interval calculator file

spiderjako:
## Description

In another PR, Yordan [proposed](https://github.com/santiment/san-chain-exporter/pull/178#discussion_r1466620295) to reform my changes in that same PR a bit and I figured that I could take those changes into a separate PR

### DIFF
--- a/blockchains/eth/lib/next_interval_calculator.js
+++ b/blockchains/eth/lib/next_interval_calculator.js
@@ -1,55 +1,52 @@
-function isNewBlockAvailable(worker) {
-  return worker.lastExportedBlock < worker.lastConfirmedBlock;
-}
 /**
- * Return the next interval to be fetched.
- * NOTE: this method modifies the member variables of its argument
- *
- * @param {*} worker A worker object, member variables would be modified
- * @returns An object like so:
- * {
- *  success: Boolean,
- *  fromBlock: Integer,
- *  toBlock: Integer
- * }
+ * Returns the context in which the worker finds itself at a given moment:
+ * 
+ * 0 : Exporting blocks that are behind the last confirmed block
+ * 
+ * 1 : We've caught up to the last confirmed block. After a query to the node, we find out that there's a higher goal
+ * 
+ * 2 : We've caught up to the last confirmed block. After a query to the node, we find out that we've caught up
+ * 
+ * @param {BaseWorker} worker A worker instance, inherriting the BaseWorker class.
+ * @returns {number} A number, which points to one of the above-given scenarios
  */
-async function nextIntervalCalculator(worker) {
-  // Check if we need to ask the Node for new Head block. This is an optimization to skip this call when the exporter
-  // is behind the last seen Head anyways.
-  const firstNewBlockCheck = isNewBlockAvailable(worker);
-  if (!firstNewBlockCheck) {
-    // On the previous cycle we closed the gap to the head of the blockchain.
-    // Check if there are new blocks now.
-    const newConfirmedBlock = await worker.web3Wrapper.getBlockNumber() - worker.settings.CONFIRMATIONS;
-    if (newConfirmedBlock > worker.lastConfirmedBlock) {
-      // The Node has progressed
-      worker.lastConfirmedBlock = newConfirmedBlock;
-    }
+async function analyzeWorkerProgress(worker) {
+  if (worker.lastExportedBlock < worker.lastConfirmedBlock) return 0;
+
+  const newConfirmedBlock = await worker.web3Wrapper.getBlockNumber() - worker.settings.CONFIRMATIONS;
+  if (newConfirmedBlock > worker.lastConfirmedBlock) {
+    worker.lastConfirmedBlock = newConfirmedBlock;
+    return 1;
   }
 
-  if (firstNewBlockCheck || isNewBlockAvailable(worker)) {
-    // If data was available without asking with Node, we are catching up and should come back straight away
-    if (firstNewBlockCheck) {
-      worker.sleepTimeMsec = 0;
-    }
-    else {
-      // If data became available only after asking the Node, we are close to the Head, come back later
-      worker.sleepTimeMsec = worker.settings.LOOP_INTERVAL_CURRENT_MODE_SEC * 1000;
-    }
+  return 2;
+}
 
-    return {
-      success: true,
-      fromBlock: worker.lastExportedBlock + 1,
-      toBlock: Math.min(worker.lastExportedBlock + worker.settings.BLOCK_INTERVAL, worker.lastConfirmedBlock)
-    };
-  }
-  else {
-    // The Node has not progressed
-    worker.sleepTimeMsec = worker.settings.LOOP_INTERVAL_CURRENT_MODE_SEC * 1000;
-    return { success: false };
-  }
+/**
+ * Function for setting the work loop's sleep time, after the end of the worker's work method.
+ * For the above given 0 and 1 scenarios, we'd want no sleep, because we have to work.
+ * For 2 we'd want to sleep, because we'd have caught up completely and should back off for some time.
+ * @param {BaseWorker} worker A worker instance, inherriting the BaseWorker class.
+ * @param {number} context The scenario used for setting the sleep time
+ */
+function setWorkerSleepTime(worker, context) {
+  worker.sleepTimeMsec = (context === 2) ? worker.settings.LOOP_INTERVAL_CURRENT_MODE_SEC : 0;
+}
+
+/**
+ * Function for calculating the next interval to be used in the worker's methods for querying the node.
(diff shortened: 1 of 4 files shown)


blockchains/eth/lib/next_interval_calculator.js
```diff
@@ -1,55 +1,40 @@
...
-    return { success: false };
-  }
+/**
+ * A function that returns the appropriate interval,
+ * depending on the progress that the worker's made.
+ * @param {BaseWorker} worker A worker instance, inherriting the BaseWorker class.
+ * @returns {object} The corresponding interval
+ */
+function nextIntervalCalculator(worker) {
+  worker.sleepTimeMsec = 0;
```
YordanPavlov, line 30: Maybe better leave the setting of the timers in a single place - the `isNewBlockAvailable`. Setting it here is in effect setting it every time there is a new block.

// Advanced discussion
Do we want to set it to 0 every time there is a new block or only when, no when we were so far behind that we do not even asked the node. This is a bit tricky, let's discuss in DM.

blockchains/eth/lib/next_interval_calculator.js
```diff
@@ -1,55 +1,40 @@
...
- * NOTE: this method modifies the member variables of its argument
- *
- * @param {*} worker A worker object, member variables would be modified
- * @returns An object like so:
- * {
- *  success: Boolean,
- *  fromBlock: Integer,
- *  toBlock: Integer
- * }
+ * If the exporter's caught up, we check for a new block. We then check whether the Node
```
YordanPavlov, line 2: This comment is confusing to me. I would instead put the comments internally to describe the if-else logic, close to the code itself, how it was before. Still subjective I guess.

spiderjako: I'll try and clear up the comment a bit, but I'm not a fan of comments inside the code. If I'm unable to do that though, I'll resort to making use of them again

test/eth/next_interval_calculator.spec.js
```diff
@@ -1,189 +1,173 @@
...
+
+    // Modify the last exported block as if we have consumed this interval
+    worker.lastExportedBlock = result.toBlock;
+
+    // Mock the Node to report an old number
+    worker.web3Wrapper.blockNumber = 90;
+
+    const contextSecond = await analyzeWorkerContext(worker);
+    setWorkerSleepTime(worker, contextSecond);
+    assert.deepStrictEqual(contextSecond, 2);
```
YordanPavlov, line 117: Use the const values and not the magic numbers here

test/eth/next_interval_calculator.spec.js
```diff
@@ -1,189 +1,173 @@
...
-
-        await nextIntervalCalculator(worker);
-        assert.deepStrictEqual(worker.sleepTimeMsec, constants.LOOP_INTERVAL_CURRENT_MODE_SEC * 1000);
-    });
-
-    it('Node gets called if Node is not ahead', async function () {
-        const worker = new eth_worker.worker(constants);
-        worker.web3Wrapper = new MockIsCalledWeb3Wrapper();
+describe('nextIntervalCalculator', () => {
+  it('returns interval according to the worker progress', () => {
```
YordanPavlov, line 77: Isn't 'block interval is not exceeded' a better name for this test. The current one is a bit vague.

test/eth/next_interval_calculator.spec.js
```diff
@@ -1,189 +1,173 @@
...
+    const worker = new eth_worker.worker(constants);
+    worker.lastExportedBlock = 0;
+    worker.lastConfirmedBlock = 150;
+
+    const { fromBlock, toBlock } = nextIntervalCalculator(worker);
+    assert.deepStrictEqual(fromBlock, 1);
+    assert.deepStrictEqual(toBlock, 100);
+  });
+
+  it('returns the interval accordingly when lastConfirmedBlock is close', () => {
```
YordanPavlov, line 87: Is this the previously called: 'Exporter should not wait for full BLOCK_INTERVAL' test? Don't you find the previous more describing of exactly what is happening?

test/eth/next_interval_calculator.spec.js
```diff
@@ -1,189 +1,173 @@
...
-
-        worker.lastExportedBlock = 1;
-        worker.lastConfirmedBlock = 2;
-
-        const resultSecond = await nextIntervalCalculator(worker);
-        assert.deepStrictEqual(resultSecond.success, true);
-        assert.deepStrictEqual(worker.sleepTimeMsec, 0);
-    });
-
-    it('Node not called if Node is ahead', async function () {
```
YordanPavlov, line 117: This test seems to be missing, don't you see value in it?

test/eth/next_interval_calculator.spec.js
spiderjako: I'm not sure, the waiting bit implies something else IMO. But I'll make it more clear, I don't know why I've named them in such a mysterious fashion

test/eth/next_interval_calculator.spec.js
```diff
@@ -1,189 +1,177 @@
...
-        const resultThird = await nextIntervalCalculator(worker);
-        assert.deepStrictEqual(resultThird.success, false);
-        assert.deepStrictEqual(worker.sleepTimeMsec, constants.LOOP_INTERVAL_CURRENT_MODE_SEC * 1000);
+describe('setWorkerSleepTime', () => {
+  it('sets sleep time accordingly when we\'re behind', () => {
+    const worker = new eth_worker.worker(constants);
+    setWorkerSleepTime(worker, 0);
 
-    });
+    assert.deepStrictEqual(worker.sleepTimeMsec, 0);
```
YordanPavlov, line 62: use const values instead of 0

test/eth/next_interval_calculator.spec.js
```diff
@@ -1,189 +1,177 @@
...
+  });
 
-    it('Fetched interval should not go backwards even if saved state is invalid', async function () {
-        const worker = new eth_worker.worker(constants);
-        worker.web3Wrapper = new MockWeb3Wrapper(4);
+  it('sets sleep time accordingly when we\'ve caught up to lastConfirmedBlock, but not the blockchain node head', () => {
+    const worker = new eth_worker.worker(constants);
+    setWorkerSleepTime(worker, 1);
```
YordanPavlov, line 67: const value

test/eth/next_interval_calculator.spec.js
```diff
@@ -1,189 +1,177 @@
...
-        worker.lastConfirmedBlock = 5;
+    assert.deepStrictEqual(worker.sleepTimeMsec, worker.settings.LOOP_INTERVAL_CURRENT_MODE_SEC * 1000);
+  });
 
-        const resultSecond = await nextIntervalCalculator(worker);
-        assert.deepStrictEqual(resultSecond.success, false);
-        assert.deepStrictEqual(worker.sleepTimeMsec, constants.LOOP_INTERVAL_CURRENT_MODE_SEC * 1000);
+  it('sets sleep time accordingly when we\'ve caught up', () => {
+    const worker = new eth_worker.worker(constants);
+    setWorkerSleepTime(worker, 2);
```
YordanPavlov, line 74: const value

test/eth/next_interval_calculator.spec.js
```diff
@@ -1,189 +1,177 @@
...
-        await nextIntervalCalculator(worker);
-        assert.deepStrictEqual(worker.web3Wrapper.isCalled, false);
-    });
+    const context = await analyzeWorkerContext(worker);
+    assert.deepStrictEqual(context, NO_WORK_SLEEP);
+  });
 
-    it('Block interval is not exceeded', async function () {
-        const worker = new eth_worker.worker(constants);
+  it('interval is correct if the blockchain node is ahead', async () => {
```
YordanPavlov, line 133: This test seems to cover functionality that has previously been tested separately in:
```
it('returns "work no sleep" case', async ()
```
and then
```
it('would not return full BLOCK_INTERVAL', ()
```

no need to have it if it adds nothing new

test/eth/next_interval_calculator.spec.js
```diff
@@ -1,189 +1,177 @@
...
-        await nextIntervalCalculator(worker);
-        assert.deepStrictEqual(worker.sleepTimeMsec, constants.LOOP_INTERVAL_CURRENT_MODE_SEC * 1000);
-    });
+  it('node is not called', async () => {
+    const worker = new eth_worker.worker(constants);
+    worker.web3Wrapper = new MockWeb3Wrapper();
+    worker.lastExportedBlock = 1;
+    worker.lastConfirmedBlock = 2;
 
-    it('Node gets called if Node is not ahead', async function () {
```
YordanPavlov, line 167: This test is similar to the one below and tests implementation. Maybe it makes sense to restore it.

YordanPavlov (approved):

PR #184, opened 2024-02-22
Migrate project to TypeScript build. Post 'signal record' on UTXO partitions.

YordanPavlov:
In this PR:
* Migrate build to TypeScript. Only some of the files are actually changed from .js to .ts. The transition would be an iterative work. However the build is in place now.
* Configurable addition of a 'signal record' to the partitions who does not receive a payload. This is useful to instruct downstream consumers that an event has taken place. Rationale is here: https://www.notion.so/santiment/Reduce-lag-on-UTXO-balances-Flink-job-3fee024bc57d42aeb380c80385bb3871

What changes in the project structure:
* `/src` directory is now where the source Typescript files live
* `/built` directory is where the compiled JS files live

I suggest the migration to happen gradually, whenever we need to edit some JS file, we would also transfer it to TS.


bin/run_cardano.sh
```diff
@@ -1,5 +1,5 @@
 #! /bin/sh
 set -e
 
-BLOCKCHAIN=cardano docker-compose -f docker/docker-compose.yaml up --build
-docker-compose -f docker/docker-compose.yaml rm -f
+BLOCKCHAIN=cardano docker compose -f docker/docker-compose.yaml up --build
```
YordanPavlov, line 4: In new versions of docker compose `docker-compose` changes to `docker compose`.

copy-assets.js
```diff
@@ -0,0 +1,18 @@
+const fs = require('fs-extra');
```
YordanPavlov, line 1: Use this script to copy .csv and .json files from the `./src` directory to the TS built directory `./built`. This is needed because those files would be left behind from the TS compilation.

docker/Dockerfile
```diff
@@ -14,10 +14,15 @@ RUN apt-get update; \
...
 WORKDIR /opt
 
-COPY package.json package-lock.json* ./
+COPY package.json package-lock.json* tsconfig.json copy-assets.js ./
+COPY src ./src
 
-RUN npm ci --${NODE_ENV}
-RUN npm cache clean --force
+RUN if [ "$BUILD_ENV" = "production" ]; \
+    then npm ci && npm prune --production; \
```
YordanPavlov, line 21: What we used before: `npm ci --${NODE_ENV}` has actually been wrong, and the env was ignored.

src/blockchains/utxo/utxo_worker.ts
```diff
@@ -1,16 +1,28 @@
 'use strict';
-const { logger } = require('../../lib/logger');
-const { constructRPCClient } = require('../../lib/http_client');
-const BaseWorker = require('../../lib/worker_base');
+import jayson from 'jayson/promise';
```
YordanPavlov, line 2: One of the file files actually migrated to TS

src/lib/kafka_storage.ts
```diff
@@ -287,29 +306,41 @@ class Exporter {
...
+      this.producer.flush(KAFKA_FLUSH_TIMEOUT, (err: LibrdKafkaError) => {
         if (err) return reject(err);
-        resolve(result);
+        resolve();
       })
     );
   }
 
-  async sendDataWithKey(events, keyField) {
+  async sendDataWithKey(events, keyField, signalRecordData = null) {
```
YordanPavlov: There is the new logic in this PR. New argument here, to be sent to the partitions which do not get a payload.

tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "compilerOptions": {
+    "outDir": "./built",
+    "allowJs": true,
+    "target": "es6",
+    "moduleResolution": "node16",
+    "allowSyntheticDefaultImports": true,
+    "sourceMap": true,
+    "esModuleInterop": true,
+    "module": "node16", // Imporant. No other value really works for us. 'commonjs' does not play well with dynamic import(). 'esnext' pretty much require to migrate everything
```
YordanPavlov, line 10: I was about to give up on the whole migration and then found this module setting. The two most common ones used are 'esnext' and 'commonjs'. Using the former pretty much requires you to migrate everything to `.ts` files and the project to ESM, which would have turned the PR into a massive undertaking. Using the latter does not work because we depend on ESM only packages (p-queue and p-retry) and they don't work in this setup.

copy_assets.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+# Define source and destination directories
+srcDir="src"
+destDir="built"
+
+# Function to copy .csv and .json files
+copyFiltered() {
```
YordanPavlov, line 8: This files moves some files which we need runtime from the `src` directory to the `built` directory. Those are `*.csv` and `*.json` files which the Typescript compiler would ignore.

src/blockchains/utxo/utxo_worker.ts
```diff
@@ -26,9 +38,10 @@ class UtxoWorker extends BaseWorker {
     });
   }
 
-  async init() {
+  async init(exporter: Exporter) {
     const blockchainInfo = await this.sendRequestWithRetry('getblockchaininfo', []);
     this.lastConfirmedBlock = blockchainInfo.blocks - this.CONFIRMATIONS;
+    await exporter.initPartitioner((event) => event['height']);
```
YordanPavlov, line 44: We are utilizing the partitioner logic for the UTXO blockchain. Records would be distributed per partition depending on block number. This should give us even distribution and also determinism where each record goes.

src/lib/kafka_storage.ts
```diff
@@ -1,59 +1,59 @@
...
+const ZOOKEEPER_SESSION_TIMEOUT: number = parseInt(process.env.ZOOKEEPER_SESSION_TIMEOUT || '30000');
+
+const FORMAT_HEADER: string = 'format=json;';
+const RDKAFKA_DEBUG: string | null = process.env.RDKAFKA_DEBUG || null;
+const KAFKA_URL: string = process.env.KAFKA_URL || 'localhost:9092';
+const KAFKA_COMPRESSION_CODEC: string = process.env.KAFKA_COMPRESSION_CODEC || 'lz4';
+const KAFKA_FLUSH_TIMEOUT: number = parseInt(process.env.KAFKA_FLUSH_TIMEOUT || '10000');
+const BUFFERING_MAX_MESSAGES: number = parseInt(process.env.BUFFERING_MAX_MESSAGES || '150000');
+const TRANSACTIONS_TIMEOUT_MS: number = parseInt(process.env.TRANSACTIONS_TIMEOUT_MS || '60000');
+const KAFKA_MESSAGE_MAX_BYTES: number = parseInt(process.env.KAFKA_MESSAGE_MAX_BYTES || '10485760');
```
WonderBeat, line 20: I'm not sure that explicit type definitions are required here if this type is private (you don't export it)

package.json
```diff
@@ -4,10 +4,12 @@
   "description": "Exporter of all transfer events for a blockchain",
   "main": "index.js",
   "scripts": {
-    "test": "TEST_ENV=1 BLOCKCHAIN=eth CONTRACT_MAPPING_FILE_PATH=test/erc20/contract_mapping/contract_mapping.json LOG_LEVEL=error CONTRACT_MODE=extract_exact_overwrite mocha --recursive --reporter spec",
+    "test": "TEST_ENV=1 BLOCKCHAIN=eth CONTRACT_MAPPING_FILE_PATH=test/erc20/contract_mapping/contract_mapping.json LOG_LEVEL=error CONTRACT_MODE=extract_exact_overwrite mocha --require source-map-support/register --recursive --reporter spec ./built/test/**/*.js",
```
WonderBeat, line 7: we have tests and app compilation result in the same `built` dir and probably add all tests to the docker distribution which is extra payload

copy_assets.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
```
WonderBeat, line 2: `set -eo pipefail` to fail fast

copy_assets.sh
```diff
@@ -0,0 +1,43 @@
...
+  local srcDir="$1"
+  local destDir="$2"
+
+  # Create destination directory if it does not exist
+  mkdir -p "$destDir"
+
+  # Read all files and directories within current directory
+  local filesAndDirs=("$srcDir"/*)
+
+  for path in "${filesAndDirs[@]}"; do
```
WonderBeat, line 18: I guess this can this be replaced by `find ... -exec ...` with less lines of code and improved readability?

docker/Dockerfile
WonderBeat: I believe `prod` and `non prod` build should be the same except that in prod build we push image to the remote repo. This shenanigans make build complex and me sad

tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "noImplicitReturns": true
+  },
+  "include": ["./src/**/*"],
+  "exclude": [
+    "node_modules",
+    "built"
```
WonderBeat, line 16: to make TS strict use

```
strictNullChecks: true,
alwaysStrict: true,
strict: true
```

and let's enable incremental!
```
incremental: true
```

WonderBeat (approved):

PR #202, opened 2024-07-26
Extend ERC20 transfers with exact balance information

YordanPavlov:
Use Mulitcall to extend ERC20 Transfer events with additional information of the exact balance of the 'to' and 'from' address at this block. This data would then be used by the Flink job.


src/blockchains/erc20/erc20_types.ts
```diff
@@ -10,5 +10,7 @@ export type ERC20Transfer = {
   logIndex: number
   type?: string,
   primaryKey?: number
+  fromBalance?: string,
```
YordanPavlov, line 13: The idea is to provide also those two fields to the Transfer data and let Flink use it

PR #204, opened 2024-09-09
Multicall fixes

YordanPavlov:
The Multicall [implementation](https://github.com/santiment/san-chain-exporter/pull/202) has been catching up on historic data up until it hit a problem where we got RPC response too big error. Some get balance responses would be too big and fail on default Ethereum Node setup with error about RPC response size being too big. Those responses look to be like either error or malicious, I think we can skip those.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,7 +9,6 @@ slaveTemplates.dockerTemplate { label ->
     container('docker') {
       def scmVars = checkout scm
       def imageName = "san-chain-exporter"
-      def ghcrRegistry = "ghcr.io"
 
       stage('Run tests') {
         sh "./bin/test.sh"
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -18,22 +17,14 @@ slaveTemplates.dockerTemplate { label ->
...
-
-          sh "docker build \
-            -t ${awsRegistry}/${imageName}:${imageVersion} \
-            -t ${awsRegistry}/${imageName}:${scmVars.GIT_COMMIT} \
-            -t ${ghcrRegistry}/santiment/${imageName}:${imageVersion} \
-            -f docker/Dockerfile ."
-
           docker.withRegistry("https://${awsRegistry}", "ecr:eu-central-1:ecr-credentials") {
-            sh "docker push ${awsRegistry}/${imageName}:${imageVersion}"
+            sh "docker build \
```
YordanPavlov, line 21: Not related with the main PR but we started getting Jenkins build errors. The issue is that we are using Tzanko's personal credentials to push the image to public Github image repository. I think we can stop doing this until / if there is some demand for this.

src/blockchains/erc20/lib/add_balances.ts
```diff
@@ -28,66 +28,112 @@ async function doMulticall(web3: Web3, blockNumber: number, addressContract: Uti
...
+   *     "data": "0xbce38bd70000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000020000000000000000000000000e2975f6a95735bdcad5e39296d37cab6470dacc80000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000002470a082310000000000000000000000007ac9368773c8d985c785ddda1964fa7c082cf89b00000000000000000000000000000000000000000000000000000000"
+   *   },
+   *     "0xd70c5d"
+   *   ]
+   *   }'
+   *
+   * This fails on our endpoint but succeeds on Ankr due to different limits. However the response is a long sequence
+   * of 0s.
+   */
+  if (errors.length > 0) {
```
YordanPavlov, line 80: Main change in PR, what we did before is a:
1. Try batch multicall.
2. On failure - try get balance per address, on failure - exit.
What we do now is:
1. Try batch multicall.
2. On failure - log warning and continue.

PR #207, opened 2024-10-25
Add ICP exporter

antisaa:


src/blockchains/icp/lib/constants.ts
```diff
@@ -0,0 +1,10 @@
+import { getIntEnvVariable } from '../../../lib/utils';
+
+export const BURN_ADDRESS = 'burn';
+export const CONFIRMATIONS = getIntEnvVariable('CONFIRMATIONS', 10);
+export const NODE_URL = process.env.NODE_URL || 'http://localhost:8545/';
+export const LOOP_INTERVAL_CURRENT_MODE_SEC = getIntEnvVariable('LOOP_INTERVAL_CURRENT_MODE_SEC', 10);
+exports.RPC_USERNAME = process.env.RPC_USERNAME;
```
YordanPavlov, line 7: Mixing two different syntaxes here, most probably this can be unified to:
```
export RPC_USERNAME = ...
```

src/blockchains/icp/icp_worker.ts
```diff
@@ -0,0 +1,211 @@
...
+            throw new Error(`Failed to fetch block after ${retries} retries: Unknown error`);
+          }
+        }
+      }
+    };
+
+    return fetchWithRetry(0); // Start with the first attempt
+}
+
+  async getTransactions(blocks: ICPBlock[]) {
```
YordanPavlov, line 108: This function holds all the domain specific knowledge, it should be unit tested.

YordanPavlov: It is a pure function, so you can move it to a separate utils file and test it easily.

YordanPavlov (approved):
Please add tests. LGTM.

PR #208, opened 2024-11-05
Add missing fields to ICP exporter

antisaa:


src/blockchains/icp/icp_worker.ts
```diff
@@ -203,9 +218,25 @@ export class ICPWorker extends BaseWorker {
...
+      this.lastPrimaryKey += transactions.length;
+    }
+
     this.lastExportedBlock += blocks.length;
-    return transactions;
+    return extendedTransactions;
   }
 }
+
+export function extendTransactionsWithPrimaryKey(transactions: Transaction[], lastPrimaryKey: number): ExtendedTransaction[] {
```
YordanPavlov, line 235: This approach of setting primary key is not ideal. We use a global incrementing counter, this causes problems when we re-extract the data  and we see different order of events or different number of events. In this case primary keys between first and second re-extraction would change and we can no longer treat the records as the same in CH. Ideal case is to use some on-chain characteristics that the blockchain protocol gives us. Like event position inside transaction. Not sure if ICP has this, just a food for thought.

antisaa: There is indeed transaction index, which we could use to have deterministic primary keys. Let's merge this PR and I'll use it in icrc(icp tokens) exporter and after it is done, update icp to new version which will use them. wdys?

YordanPavlov (approved):

PR #210, opened 2024-12-04
Cardano supports block interval

YordanPavlov:
Support block interval in Cardano exporter. 

The Cardano exporter uses GraphQL to fetch transactions. It is not necessary to specify end block, as the API would return as many blocks as can fit in the response. However adding the interval allows for better debugging when we want to fetch less data than the maximum.

### DIFF
--- a/src/blockchains/cardano/cardano_worker.ts
+++ b/src/blockchains/cardano/cardano_worker.ts
@@ -5,9 +5,6 @@ import util from './lib/util';
 import { logger } from '../../lib/logger';
 
 
-const CARDANO_GRAPHQL_URL = process.env.CARDANO_GRAPHQL_URL || 'http://localhost:3100/graphql';
-const DEFAULT_TIMEOUT_MSEC = parseInt(process.env.DEFAULT_TIMEOUT || '30000');
-
 export class CardanoWorker extends BaseWorker {
   private pRetry: any;
   private got: any;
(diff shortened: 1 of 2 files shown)


src/blockchains/cardano/cardano_worker.ts
```diff
@@ -5,9 +5,6 @@ import util from './lib/util';
 import { logger } from '../../lib/logger';
 
 
-const CARDANO_GRAPHQL_URL = process.env.CARDANO_GRAPHQL_URL || 'http://localhost:3100/graphql';
```
YordanPavlov, line 8: Move those variables to where the others live. They come in the constructor argument.

src/blockchains/cardano/cardano_worker.ts
```diff
@@ -115,17 +112,18 @@ export class CardanoWorker extends BaseWorker {
...
-    const query = `
-    {
-      transactions(
-        where: {
-          block: { epoch: { number: { _is_null: false } } }
-          _and: [{ block: { number: { _gte: ${blockNumber} } } },
-                 { block: { number: { _lte: ${lastConfirmedBlock} } } }]
-        }
-        order_by: { includedAt: asc }
+  async getTransactions(fromBlock: number, toBlock: number) {
```
YordanPavlov, line 115: Fetch `from-to` blocks instead of `from-last`

src/blockchains/cardano/cardano_worker.ts
```diff
@@ -149,18 +147,24 @@ export class CardanoWorker extends BaseWorker {
...
+
+    const response = await this.pRetry(() => this.sendRequest(query), {
+      onFailedAttempt: (error: any) => {
+        // Determine the error message
+        const errorMessage = error.cause?.message || error.message || 'Unknown error';
+
+        logger.warn(
+          `Request ${fromBlock} - ${toBlock} is retried. ` +
+          `There are ${error.retriesLeft} retries left. ` +
+          `Error: ${errorMessage}`
```
YordanPavlov, line 160: Log error as part of WARN message

PR #212, opened 2024-12-13
Internal transaction position instead of primary key

YordanPavlov:
Instead of generating 'primary key' inside ETH transfers exporter, use 'internal transaction position'. That is, when two transfers match by other fields, increment the internal transaction position so to generate ordering.


src/blockchains/eth/lib/end_of_block.ts
```diff
@@ -1,22 +1,10 @@
-import { ETHBlock } from "../eth_types";
+import { ETHBlock, ETHTransfer } from "../eth_types";
 import { Web3Interface, safeCastToNumber } from "./web3_wrapper";
 
-export type EOB = {
```
YordanPavlov, line 4: Do we need the special EOB type? It seems we want it to match the ETHTransfers type structure exactly, so maybe it should not be a special type?

PR #216, opened 2025-02-14
Bugfix. ERC20 transfers events out of order.

YordanPavlov:
* When a log has 'logIndex' above 10000 the primary key generation would overflow and the primary keys would mix with next block's keys.
* In some modes we would order by primary key only and this produces wrong ordering in the Kafka topic.

### DIFF
--- a/src/blockchains/erc20/erc20_worker.ts
+++ b/src/blockchains/erc20/erc20_worker.ts
@@ -173,6 +173,10 @@ export class ERC20Worker extends BaseWorker {
     // If overwritten events have been generated, they need to be merged into the original events
     if (overwritten_events.length > 0) {
       stableSort(resultEvents, function primaryKeyOrder(a: ERC20Transfer, b: ERC20Transfer) {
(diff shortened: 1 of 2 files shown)


src/blockchains/erc20/erc20_worker.ts
```diff
@@ -173,6 +173,10 @@ export class ERC20Worker extends BaseWorker {
     // If overwritten events have been generated, they need to be merged into the original events
     if (overwritten_events.length > 0) {
       stableSort(resultEvents, function primaryKeyOrder(a: ERC20Transfer, b: ERC20Transfer) {
+        const blockDif = a.blockNumber - b.blockNumber;
```
YordanPavlov, line 176: This bug consists of two parts.
* Primary key generation is flawed. We presume that there would not be a logIndex above 10000. However when this happens, the primary key for the event with the huge log index would become bigger number than the primary key for events for the next block.
* We used to sort by primary key only.
The fix we do here is that even though primary key construction is flawed we sort by block number first.

antisaa (approved):

PR #220, opened 2025-03-06
Use verification node when doing ETH QC

YordanPavlov:
In a [previous PR](https://github.com/santiment/san-chain-exporter/pull/213) we added some quality checks to our ETH exporter. We verify that:
* No blocks are missing
* Transaction number increase with +1
In the previous review @WonderBeat raised a concern that there _are_ expected empty blocks. Me and Ante thought we should see some sort of reward on all blocks though. Now, once the exporter has been running for a while it appears there are empty blocks, so the previous assumption is wrong. 
What the current PR does is it adds a second Node which is uses only for verification if it also returns empty block data when the main node does.

Not related with the main idea of the PR I got into a bit of a refactoring. We have this Web3Wrapper class which wraps all work with the Web3 library. However for data transformation purposes (hex to decimal, etc) we do not need a connection to a node and those calls can be made static. This simplifies our code as we no longer need to provide a connection to all parts of the code.


src/blockchains/erc20/lib/fetch_events.ts
```diff
@@ -37,7 +37,7 @@ export function decodeEventBasicInfo(event: any, timestampsCache: TimestampsCach
 /**Transfer(address,address,uint256)
  * Used by all ERC20 tokens
  **/
-function decodeTransferEvent(web3Wrapper: Web3Interface, event: any, timestampsCache: TimestampsCacheInterface): ERC20Transfer | null {
+function decodeTransferEvent(event: any, timestampsCache: TimestampsCacheInterface): ERC20Transfer | null {
```
YordanPavlov, line 40: There are a lot of this type of changes. Argument is no longer provided as the function can use static instance.

src/blockchains/eth/eth_worker.ts
```diff
@@ -106,11 +112,11 @@ export class ETHWorker extends BaseWorker {
     const [traces, blocks, receipts] = await this.fetchData(fromBlock, toBlock)
     const events: (ETHTransfer | EOB)[] = this.transformPastEvents(fromBlock, toBlock, traces, blocks, receipts)
     events.sort(transactionOrder)
-    if (this.settings.CHECK_QUALITY) {
-      checkETHTransfersQuality(events, fromBlock, toBlock)
+    if (this.ethClientVerification) {
```
YordanPavlov, line 115: The main task in this PR - use verification node. Here we use the presence of a verification node URL as indicator to do QC.

src/blockchains/eth/lib/util.ts
```diff
@@ -33,27 +36,26 @@ export function assignInternalTransactionPosition(transfers: ETHTransfer[], grou
   })
 }
 
-export function assertBlocksMatch(groupedTransfers: any, fromBlock: number, toBlock: number) {
+export async function assertBlocksMatch(groupedTransfers: any, fromBlock: number, toBlock: number,
+  ethClientVerification: HTTPClientInterface) {
   const keys = Object.keys(groupedTransfers)
 
-  // A list of empty blocks, no transfers and no rewards.
-  const blocksExceptionList = [15537454]
```
YordanPavlov, line 40: I thought we can whitelist a small number of blocks. It didn't work out.

src/blockchains/eth/lib/web3_wrapper.ts
```diff
@@ -22,62 +16,68 @@ function containsOnly0Andx(input: string) {
...
+        return this.web3;
+    }
+
+}
+
+
+/**
+ * A set of static Web3 functions which do not require connection to a Node
+ */
+export class Web3Static {
```
YordanPavlov, line 47: This class now holds all Web3 functions which do not require a connection.

src/blockchains/receipts/lib/helper.ts
```diff
@@ -0,0 +1,96 @@
+const lang = require('lodash/lang');
```
YordanPavlov, line 1: File got changed from JS to TS.

antisaa (approved):

PR #225, opened 2025-04-17
Option to disable Kafka producer sticky partition

YordanPavlov:
In this PR we make it configurable to disable Kafka producer 'sticky partition'.

When we don't provide an explicit partition (and key record is null) the Kafka producer would do roundrobin. However it would also apply a 'sticky partition' concept where it batches record and send them in the same partition and only then move to the next one. This sticky behavior is intended to improve performance, however in our case we want to shuffle records as good as possible across partitions, ideally we want to have a record for a block number written in all partitions. Reason for this is downstream consumer Flink, 'sticky' behavior results in all records for a block landing in same partition, then Flink needs to wait for data to arrive in other partitions and this adds artitifical delay in our processing. Flink watermarks dashbord visualizes this problem:

image

What we see here is that some sources have new data but not all, and the global Watermark is choppy instead of smooth. Disabling sticky behavior we end up with this chart:

image

(the right part)

### DIFF
--- a/src/lib/kafka_storage.ts
+++ b/src/lib/kafka_storage.ts
@@ -43,13 +43,13 @@ class Partitioner {
 
   async init(hashFunction: ((value: object) => number), topicName: string, producer: Kafka.Producer) {
     this.hashFunction = hashFunction;
-    this.partitionCount = await this.findPartitionCount(topicName, producer);
+    this.partitionCount = await this.#findPartitionCount(topicName, producer);
     if (this.partitionCount <= 0) {
       throw new Error('Partition count should be > 0');
     }
   }
 
-  async findPartitionCount(topicName: string, producer: Kafka.Producer) {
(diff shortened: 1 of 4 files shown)


src/lib/kafka_storage.ts
```diff
@@ -296,42 +300,24 @@ export class Exporter {
     }
   }
 
-  async sendData(events: Array<any>) {
```
YordanPavlov, line 299: Unify those two methods, only one was being actively used.

src/lib/kafka_storage.ts
```diff
@@ -133,6 +133,10 @@ export class Exporter {
       producer_settings['enable.idempotence'] = true;
     }
 
+    if (disableStickyPartition) {
+      producer_settings['sticky.partitioning.linger.ms'] = 0;
```
YordanPavlov, line 137: This results in truly roundrobin behavior.

WonderBeat (approved):

PR #227, opened 2025-05-14
Internal tx position is assigned for same tx

YordanPavlov:
* Some aritifical transfers we generate have tx position set to 0 even though 'transactionHash' is unique.
* When assigning internal tx position we should consider the tx position and not the transacitonHash because this is based on what Flink would deduplicate.

The problem appears for records like this:

```
{"from":"0x109c4f2ccc82c4d77bde15f306707320294aea3f","to":"0xfd2605a2bf58fdbb90db1da55df61628b47f9e8c","value":500000000000000000,"valueExactBase36":"3sr77gzykagw","blockNumber":50324,"timestamp":1438987309,"transactionHash":"0x22591c7aadb705df30a13e97f73496b529f208573a231c7eb389766d29b421d1","transactionPosition":0,"internalTxPosition":1,"type":"call"}
{"from":"mining_block","to":"0x3f113041d56dda809a245fe338832c2d373e7c80","value":5000000000000000000,"valueExactBase36":"11zk02pzlmwow","blockNumber":50324,"timestamp":1438987309,"transactionHash":"mining_block","transactionPosition":0,"internalTxPosition":0,"type":"reward"}
```

### DIFF
--- a/src/blockchains/eth/lib/util.ts
+++ b/src/blockchains/eth/lib/util.ts
@@ -23,7 +23,7 @@ export function transactionOrder(a: ETHTransfer | EOB, b: ETHTransfer | EOB) {
   return internalTxPositionA - internalTxPositionB
 }
 
-const ethTransferKey = (transfer: ETHTransfer) => `${transfer.blockNumber}-${transfer.transactionHash ?? ''}`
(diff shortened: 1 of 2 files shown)


antisaa (approved):
LGTM

PR #251, opened 2026-07-31
ERC20: handle blocks exceeding the node eth_getLogs limit

YordanPavlov:
## Problem

The ERC20 worker crashed on Ethereum with:

```
Returned error: query returns too many logs, narrow your filter: 20000
```

even with `BLOCK_INTERVAL` reduced to a single block. Since the worker fetches **all** logs in the range (the topics filter is commented out due to an old Parity bug), any block with more than 20000 logs is unfetchable via `eth_getLogs` — block 25649104 has 21176 logs (neighbors: ~700–1300), so no block batching can get under the limit.

## Fix

On the "too many logs" error, `getRawEvents` now:

1. recursively halves the block interval;
2. when a single block is still over the limit, fetches its logs via `eth_getBlockReceipts` (no response cap) and flattens the receipt logs.

The `contractAddress` filter is applied client-side on the fallback path; topic filtering already happens downstream in `decodeEvents`, so the output is identical to the normal path. Normal operation is unchanged — the fallback only runs on this specific error, other errors are rethrown as before.

## Tests

New `fetch_events_fallback.spec.ts` covers interval splitting + receipts fallback (ordering preserved, receipts fetched once), the single-block case, contract-address filtering on the fallback path, and that unrelated errors still propagate. All 229 tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/blockchains/erc20/lib/fetch_events.ts
+++ b/src/blockchains/erc20/lib/fetch_events.ts
@@ -234,6 +234,28 @@ export async function getPastEvents(web3Wrapper: Web3Interface, fromBlock: numbe
 }
 
 
+function isTooManyLogsError(error: any): boolean {
+  const message = String(error?.message ?? error);
+  return /too many logs|more than \d+ (results|logs)/i.test(message);
+}
+
+// Fetch all logs of a single block through eth_getBlockReceipts, which is not subject
+// to the node's eth_getLogs response limit.
+async function getRawEventsFromReceipts(web3Wrapper: Web3Interface, blockNumber: number,
+  contractAddress: string | string[] | null): Promise<RawEvent[]> {
+  const receipts = await web3Wrapper.getBlockReceipts(blockNumber);
+  let logs: RawEvent[] = receipts.flatMap((receipt: any) => receipt.logs ?? []);
+
+  if (contractAddress) {
+    const allowed = new Set(
+      (Array.isArray(contractAddress) ? contractAddress : [contractAddress]).map(a => a.toLowerCase())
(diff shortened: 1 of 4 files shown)


src/blockchains/erc20/lib/fetch_events.ts
```diff
@@ -234,6 +234,28 @@ export async function getPastEvents(web3Wrapper: Web3Interface, fromBlock: numbe
 }
 
 
+function isTooManyLogsError(error: any): boolean {
+  const message = String(error?.message ?? error);
+  return /too many logs|more than \d+ (results|logs)/i.test(message);
```
YordanPavlov, line 239: Figure out if the error returned by the Node is related to too many logs

src/blockchains/erc20/lib/fetch_events.ts
```diff
@@ -234,6 +234,28 @@ export async function getPastEvents(web3Wrapper: Web3Interface, fromBlock: numbe
...
+}
+
+// Fetch all logs of a single block through eth_getBlockReceipts, which is not subject
+// to the node's eth_getLogs response limit.
+async function getRawEventsFromReceipts(web3Wrapper: Web3Interface, blockNumber: number,
```
YordanPavlov, line 244: Fallback to this method when a lot of logs are coming from a single block. We do not use it for all extract as it do not support a range of blocks.

antisaa (approved):

apozdeyev (commented):
looks good!

PR #252, opened 2026-07-31
Multicall blacklist added

YordanPavlov:
A multicall blacklist. When a contract fails a multicall several times - blacklist it to unload the API load and also improve on logging.

Our logs were full of a small set of contracts failing the multicall `balanceOf` we do. So decided to both minimize the logging and reduce the api load be excluding those through a blacklist.

The blacklist is not persisted across restarts. Consider this implementation but a) it's a bit more complex b) it's unclear under what conditions we want to allow for rehabilitation. Maybe a contract got updated and do not fail anymore.

### DIFF
--- a/src/blockchains/erc20/lib/add_balances.ts
+++ b/src/blockchains/erc20/lib/add_balances.ts
@@ -6,6 +6,7 @@ import { MULTICALL_ADDRESS } from './constants';
 import * as Utils from './balance_utils';
 import { logger } from '../../../lib/logger';
 import { buildInclusiveChunks } from './chunk_utils';
+import { MulticallBlacklist } from './multicall_blacklist';
 
 const stringifyTransfer = (event: ERC20Transfer): string =>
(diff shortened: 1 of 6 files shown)


src/blockchains/erc20/lib/add_balances.ts
```diff
@@ -115,33 +147,84 @@ async function executeBatchMulticall(web3: Web3, addressContracts: Utils.Address
...
         return decodeMulticallResult(rawResult, web3, blockNumber)
       }
     })
+}
+
+// Feed the per-contract outcomes to the blacklist.
+function recordBlacklistSignals(stats: ContractStats, blockNumber: number, blacklist: MulticallBlacklist) {
+  for (const [contract, contractStats] of Array.from(stats.entries())) {
+    if (contractStats.failed < contractStats.total) {
+      blacklist.recordCleanResult(contract);
```
YordanPavlov, line 202: Success is recorded because we would blacklist only if contract fails always. Our trigger is a number of consecutive failures without success among them.

WonderBeat (approved):

